PR #8, opened 2018-06-05
Optional title

### LINKED ISSUE #2: Make title optional
The aim of this editor isn't to be incredibly flexible, if you want that use Slate – however one small change would make this useful for a much wider array of usecases, make the forced title section optional.

jorilallo:
With `title={false}` title block isn't required anymore.

Fixes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ See a working example in the [example directory](/example).
 A markdown string that represents the initial value of the editor. Use this to prop to restore
 previously saved content for the user to continue editing.
 
+#### `title`
+
+With `title` set to `false` the editor won't require title at the beginning of the document. The default is `true`.
+
 #### `titlePlaceholder`
 
 Allows overriding of the placeholder text displayed where a title would be. The default is "Your title".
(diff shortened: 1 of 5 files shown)


tommoor:
Thanks for taking a look at this, it's definitely made me ponder on the right way to implement this and it'd be good to get your feedback. A couple of thoughts:

- I'm tempted to just pull the title block entirely out of `rich-markdown-editor` and make it a separate plugin in `Outline` to simplify the code and better match Slate's approach to functionality.

- If we don't do that then the enforcing of the title should probably be moved to a plugin within the editor, then passing `title=true` could simply add that plugin to the stack instead of massaging the default schema (plugins can provide their own schema rules)

I also noticed that `schemas` now have a [first rule](https://docs.slatejs.org/slate-core/schema#first) that might be a better way to force the title block, not sure though.

jorilallo:
Plugin would be better, this definitely felt hacky and not clear at all. I
also think that title off by default would be a better default. Feel free
to close, I can use my fork in the meantime.

PR #17, opened 2018-07-13
Dist improvements

jorilallo:
Also `transform-async-to-generator` requires regenerator runtime which I missed somehow :( Added runtime plugin to include it and also swapped to using es5 build of the boundless library (it exposes `module` in the `package.json` so non-standard ES6 was used by default). Tested with CRA.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,6 +7,12 @@
     "transform-es2015-destructuring",
     "transform-object-rest-spread",
     "transform-async-to-generator",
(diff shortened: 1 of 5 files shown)


jorilallo:
I guess we ran into this before but as we're not using webpack for builds, not sure if there's another way to resolve boundless import issue https://github.com/outline/outline/pull/507#discussion_r160029464

PR #21, opened 2018-07-20
Add missing uuid dependency

sanohin:


tommoor:
Wow, I wonder how this was even building 🤔 

Considering this is only used in one spot I might just remove the dep altogether and just use a timestamp or something for the temporary upload id

sanohin:
@tommoor you can try this https://github.com/ai/nanoid

tommoor:
Thanks for bringing it to attention @sanohin – it's literally used as just a temporary DOM id so I went ahead with [using an incrementing number](https://github.com/outline/rich-markdown-editor/commit/d7ca4fa7ec62a770e9253ba0be928dd92d778cdb) stead. Having a uuid was just a hold-over from when this code was taken out of the main Outline codebase 😄

PR #23, opened 2018-08-01
Convert this library to emotion

jfrolich:
Not sure if you're up to it. But I converted the library to emotion (basically a drop-in replacement).

Emotion is a bit better to use for external packages than styled components because the size of the dependency is smaller. React-select also went this route.

It would even better if we could refactor out the styled components API, and do theming as is done in react-select (functions that yield styles for particular elements). In that case the difference is a gzip size of 5kb vs 15kb for styled components. 

Additionally there is a large performance benefit vs styled components (see https://github.com/A-gambit/CSS-IN-JS-Benchmarks/blob/master/RESULT.md).

Let me know what you think!


jorilallo:
As Outline already uses Styled Components, it would probably make more sense to use RME with a shim and alias, although Emotion doesn't like `styled.div` syntax without the babel plugin (I tested this quick). Maybe it makes more sense to have a `emotion` based fork...

jfrolich:
I'm ok maintaining an emotion based fork. 

What would be an option perhaps is have the library be based on emotion (so it can use the babel plugin), and then use an alias if you want to use styled components with it (not sure what RME means).

Also if we remove the styled component API from the editor, it's just a 5kb overhead for emotion, so the impact would not be huge if the adopting app would use styled components for other components. It would also allow for a nice API to customize the styles that I think is easier to accomplish with emotion than styled-components (`react-select` provides a nice reference how that can be done).

jorilallo:
@tommoor I used it with Linear, but in a pretty SC way. Nothing wrong to say about it really apart from the annoying Typescript churn recently (haven't updated from 2.8 yet...). Lighter size is nice but SC is more popular I guess. Kind of up to you, happy to help but if we set up a fork, I might move to it as we're trying to strict about the bundle size. My preference might be slightly on emotion.

@jfrolich Curious if you're in the TS camp or not.

tommoor:
I'm closing this for now, this library will remain using styled components but feel free to maintain an emotion fork of course.

jfrolich:
@jorilallo: I'm using Flow in the main app I am working on, but Typescript has come a long way since making the decision to go with Flow. 

I tried to use Typescript some time ago, and found it annoying/challenging to type everything. I'd like to type everything lightly, and only 100% type important parts of the app, which used to work slightly better with Flow, but that might be better now.

The dev tooling and library support though is soo much better with Typescript, so definitely thinking about converting the app sometime.

I will just maintain an Emotion fork for the folks that use Emotion, so it's not a big overhead, it's a small diff anyway. Happy to help think about making it more efficient as a library later, but this will work for now 👍.

PR #34, opened 2018-08-15
Autofocus and optional methods

jorilallo:
- Added `autoFocus` so that the editor can nicely sit in a form with other fields
- Made `onSave` and `onCancel` optional in case the host application wants to disregard or implement themselves


jorilallo:
@tommoor yeah, I guess it would be. Couple of tweaks to this:

- Right now, if the editor has content on mount, it won't autoFocus. Cool if I change it so that it will, feel more uniform that way?
- Flow is being stupid about the optional props, right? Or am I stupid
- I'll also make `onImageUploadStart` and `onImageUploadStop` optional as they shouldn't be required if `uploadImage` is omitted

tommoor:
I think it's okay to change that autoFocus with content - puts the choice in the host apps hands. optional ALL THE THINGS.

tommoor:
I'd say flow is being funky - try upgrading? It's 10 versions behind in this repo.

jorilallo:
@tommoor ready for 👀

src/index.js
```diff
@@ -90,8 +91,8 @@ class RichMarkdownEditor extends React.PureComponent<Props, State> {
     if (this.props.readOnly) return;
     window.addEventListener("keydown", this.handleKeyDown);
 
-    if (!this.props.defaultValue) {
-      this.focusAtStart();
+    if (this.props.autoFocus) {
+      this.focusAtEnd();
```
tommoor: Ah man, I missed that you changed this from `focusAtStart` to `focusAtEnd` ... I'm going to change this in 5.0.1 to focus at the start if `autoFocus && !this.props.defaultValue` otherwise it really screws up Outline which has an empty title and paragraph by default

cc @jorilallo

PR #35, opened 2018-08-26
Hidden toolbar buttons

jfrolich:
This PR will allow users to selectively turn off toolbar buttons in the theme.

I also created a type definition for the theme.

### DIFF
--- a/src/components/Toolbar/BlockToolbar.js
+++ b/src/components/Toolbar/BlockToolbar.js
@@ -16,14 +16,14 @@ import {
 } from "outline-icons";
 import getDataTransferFiles from "../../lib/getDataTransferFiles";
 import Flex from "../Flex";
-import type { SlateNodeProps } from "../../types";
+import type { SlateNodeProps, Theme } from "../../types";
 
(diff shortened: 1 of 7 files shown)


tommoor:
Would you mind removing the extensive `yarn.lock` update from this PR before I merge?

jorilallo:
@tommoor passing arrays of formatting options would be cool. I was thinking of adding this as I want to edit mine and remove the block toolbar

jfrolich:
Yeah open to API/naming feedback. I needed to remove H1 in my app. Happy to do some polish.

I will remove the yarn.lock 👍

jfrolich:
brought back yarn.lock from master (so no diff). BTW it would be great to fix the issue with node 10 in yarn.lock!

tommoor:
Thanks!

@jorilallo it seems reasonable to merge this as a non breaking change and then if someone wants to build out passing arrays of formatting options that can succeed this API at that time.

jorilallo:
Yeah I might look into it once it's more timely for myself. But good to know you're open to it :)

PR #36, opened 2018-08-26
Remove unnecessary componentWillUpdate

jfrolich:
This lifecycle method will not be supported anymore by async react

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -52,7 +52,6 @@ type Props = {
 type State = {
   editorValue: Value,
   editorLoaded: boolean,
-  schema: Schema,
 };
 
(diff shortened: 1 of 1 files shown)


tommoor:
Thanks for the PR.

I think it's best that when it's deprecated all the call sites are moved to `getDerivedStateFromProps` or remove the usage of state altogether, as it stands today this doesn't provide any benefits and running additional checks inside of `render` definitely isn't best practice either.

tommoor:
> It might be more declarative to have a this.getSchema() with caching instead of the more imperative this.updateSchema() in terms of naming :)

I can get behind that 👍

PR #46, opened 2018-10-22
remove image after failed upload

defunctzombie:
Remove images which fail to upload from the document. An image which
fails to upload will not be available on subsequent document loading and
thus should not continue to appear in the document if it fails to
upload. Keeping the image in the document is misleading to the user
(even with the "sorry failed to upload" overlay).

This change removes the image from the document and shows a toast
message instead.

---
See https://github.com/outline/outline/issues/786 for more context.

### DIFF
--- a/src/changes.js
+++ b/src/changes.js
@@ -85,9 +85,11 @@ export async function insertImageFile(
         data: { src, alt, loading: false },
       };
     } catch (error) {
-      props = {
-        data: { alt, src: placeholderSrc, loading: false, error },
-      };
+      if (editor.props.onShowToast) {
+        editor.props.onShowToast(
+          "Sorry, an error occurred uploading the image"
+        );
+      }
     }
 
     const placeholder = editor.value.document.findDescendant(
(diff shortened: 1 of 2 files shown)


defunctzombie:
@tommoor I've taken a first draft pass at the functionality - verified using the repo's example. Need a few high level thoughts on this to polish it off.

src/changes.js
```diff
@@ -98,6 +98,21 @@ export async function insertImageFile(
...
+    if (props.data.error) {
+      editor.change(change => {
+        change.removeNodeByKey(placeholder.key);
+      });
+
+      if (editor.props.onShowToast) {
+        editor.props.onShowToast("Sorry, an error occurred uploading the image")
+      }
+
+      return;
```
tommoor: I think we can remove the return and use `else` instead. No need to call `onImageUploadStop` separately – it will always be called within the finally block

defunctzombie:
@tommoor updated with suggestion and final TODO items. PTAL.

PR #51, opened 2018-12-20
Retina images

jorilallo:
Implemented some PNG inspection code to determine if uploaded images are retina screenshots to scale them properly. This was hard enough so didn't bother to start looking at the blob stage size but that would be an improvement.

### DIFF
--- a/src/components/Image.js
+++ b/src/components/Image.js
@@ -2,8 +2,8 @@
 import * as React from "react";
 import ImageZoom from "react-medium-image-zoom";
 import styled from "styled-components";
-import extract from "png-chunks-extract";
 import type { SlateNodeProps as Props } from "../types";
+import { isRetinaImage } from "../lib/imageHelpers";
 
 type State = {
   hasError?: boolean,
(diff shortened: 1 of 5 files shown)


src/components/Image.js
```diff
@@ -2,14 +2,64 @@
...
     hasError: false,
+    imageWidth: undefined,
+  };
+
+  /**
+   * Detect if PNG image is in retina resolution.
+   *
+   * PNG Chunk documentation: https://www.w3.org/TR/PNG-Chunks.html
+   */
+  onImgLoad = async (ev: SyntheticEvent<HTMLImageElement>) => {
```
tommoor: I guess you have no desire to pull this into it's own module? I'd probably move into it's own lib file at least.

Will test thoroughly this w/e.

jorilallo: Moved to a lib file. Not particularly fond of publishing 10 lines as a module :)

tommoor:
Unfortunately the PHYs chunk seems to be missing when copying an image straight to your clipboard (hold shift while taking screenshot) 😢 – not sure why that would be, but saving to disk first works.

src/components/Image.js
```diff
@@ -3,13 +3,28 @@ import * as React from "react";
...
   state = {
     hasError: false,
+    imageWidth: undefined,
+  };
+
+  onImgLoad = async (ev: SyntheticEvent<HTMLImageElement>) => {
+    const image = ev.target;
+    const url = this.props.node.data.get("src");
+    // Skip upload stage
+    if (url.startsWith("blob:")) return;
```
tommoor: In my testing you don't need this line at all, the browser can fetch a blob uri without issue

tommoor:
It seems like there's always a flash of the image at full size causing all the layout to shift, have you seen that in your testing?

PR #62, opened 2019-02-14
Make inbuilt renderNode a plugin

gscoon:
**Goal:**
Provide the ability to create a custom renderNode with **priority** for specific node types, but then fallback to the inbuilt node for all others.

The current setup forces you to either: 
a) completely override the inbuilt renderNode   
b) have a custom plugin-based renderNode that's secondary to the inbuilt

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ import { history } from "react-router";
 
 #### `renderNode`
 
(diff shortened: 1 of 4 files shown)


tommoor:
Nice, this looks like a good improvement! I'll give it a test soon, but don't forsee any issues merging – as it's a breaking change it will be published as a major release.

PR #66, opened 2019-02-15
Adds `jpeg` MIME type to the configured extensions image insert.

johnmcdowall:
Chrome 72 on OSX is reporting the type of dropped .jpg files as `jpeg`,
which means they are not accepted when they should be. I add the `jpeg`
MIME type along with the existing `jpg` type so that all JPG files are
accepted.


tommoor:
Many thanks, will publish in a new release shortly.

Issue #71, opened 2019-03-12
Creating a link from an empty selection triggers an error


glaand:
At the canvas if you double-click at an empty starting line and select the link option, it triggers an error.

<img width="1429" alt="grafik" src="https://user-images.githubusercontent.com/17197309/54240692-116a6980-451f-11e9-9d72-f457f3a7d152.png">

The error is triggered by a TypeError
<img width="858" alt="grafik" src="https://user-images.githubusercontent.com/17197309/54240721-247d3980-451f-11e9-8a7d-46f01b1458c1.png">

tommoor:
closed by https://github.com/outline/rich-markdown-editor/commit/3bc474b4b3dd376cb650b889cd71d2b761afd6ce

PR #72, opened 2019-03-12
Fix Issue #915 (outline/outline): Creating a link from an empty selection triggers an e…

glaand:
#71 Fixed!

### DIFF
--- a/src/components/Toolbar/FormattingToolbar.js
+++ b/src/components/Toolbar/FormattingToolbar.js
@@ -75,9 +75,14 @@ class FormattingToolbar extends React.Component<Props> {
     ev.preventDefault();
     ev.stopPropagation();
 
-    const data = { href: "" };
-    this.props.editor.wrapInline({ type: "link", data });
-    this.props.onCreateLink(ev);
+    let selection = window.getSelection().toString();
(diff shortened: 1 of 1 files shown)


src/components/Toolbar/FormattingToolbar.js
```diff
@@ -77,7 +77,9 @@ class FormattingToolbar extends React.Component<Props> {
 
     const data = { href: "" };
     this.props.editor.wrapInline({ type: "link", data });
-    this.props.onCreateLink(ev);
+    if (data.href.length) {
```
tommoor: You're checking a value that is set to an empty string above on L78  so this callback will never be triggered I'm afraid.

tommoor: It might be that we want to check the length of the selection before ever calling `handleCreateLink`? 🤔

glaand: Right haha, i'm not used to react but i managed to get the selection.  I've already pushed my new commit, what do you think? Makes the calling of the trim function sense?

tommoor:
This does the trick, thanks for your contribution 💯

Issue #79, opened 2019-04-22
Server side rendering


WNemencha:
It seems that currently this component doesn't supports SSR. We are using Next.JS internaly and would like to use this to display our blog posts, but we can't only rely on client-side rendering because that would defeat the whole point of using Next, that is SEO for free.

I will push a PR with SSR support soon, this issue is here to keep track of progress !

tommoor:
Hey William, great! Looking forward to receiving your PR for this

WNemencha:
@tommoor I've just pushed a PR, feel free to tell me if something needs to be changed to match the contributions guidelines or stuff. :stuck_out_tongue:

kelvien:
Does SSR here mean that we can render the content in `defaultValue`/`value` props?

In the PR I only see changes within the `componentDidMount`, and if I understand SSR correctly in React, `componentDidMount` is not a part of a lifecycle that will be executed in SSR (`React.renderToString`). Can someone explain to me how the PR above enables SSR?

I am also trying to render my editor's content in the server-side, but it doesn't seem like it's supported yet. And I'm using rich-markdown-editor version `10.4.0`

tommoor:
SSR just means that the component can be rendered on a server (no window object).

I'm not sure how it performs with rehydration - it's possible other fixes are needed, this isn't a usecase I've tested against personally.

kelvien:
Got it. That is what I would expect too, but it doesn't seem like the Editor renders the initial values on server because the values initialization is done during `componentDidUpdate` which in SSR case, that lifecycle does not exist on server-side.

kelvien:
Ah I think I see what's going on, I looked at the state of the Editor file on this [commit](https://github.com/WNemencha/rich-markdown-editor/blob/1cd3580df4c110f9a9f495fb9ccd0938c02e40d4/src/index.js#L79)
The value of the `StyledEditor` is defined in the [constructor](https://github.com/WNemencha/rich-markdown-editor/blob/2bef93037fee62d9ec0b2df2fb08d113ae5c83f1/src/index.js#L79)
Whereas in the latest version the value is defined in the `componentDidMount` hence it's never rendered in the server

PR #80, opened 2019-04-23
feat(ssr-support): Check existence of window before using it so no errors are thrown in SSR pages

WNemencha:
Fix issue #79. That's the only changes needed to make this working with server side rendering. At least with Next. I could have checked for `process.browser` but I'm unsure this is well supported so `typeof` is usually the best way to achieve this.

### DIFF
--- a/src/components/BlockInsert.js
+++ b/src/components/BlockInsert.js
@@ -44,12 +44,12 @@ class BlockInsert extends React.Component<Props, State> {
   };
 
   componentDidMount = () => {
-    window.addEventListener("mousemove", this.handleMouseMove);
+    typeof window != 'undefined' && window.addEventListener("mousemove", this.handleMouseMove);
   };
 
   componentWillUnmount = () => {
     if (this.mouseMoveTimeout) clearTimeout(this.mouseMoveTimeout);
-    window.removeEventListener("mousemove", this.handleMouseMove);
+    typeof window != 'undefined' && window.removeEventListener("mousemove", this.handleMouseMove);
   };
 
   setInactive = () => {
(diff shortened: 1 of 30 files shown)


src/components/BlockInsert.js
```diff
@@ -44,12 +44,12 @@ class BlockInsert extends React.Component<Props, State> {
   };
 
   componentDidMount = () => {
-    window.addEventListener("mousemove", this.handleMouseMove);
+    typeof window != 'undefined' && window.addEventListener("mousemove", this.handleMouseMove);
```
tommoor: Please update to `!==` in this file, if you push again hopefully I've changed the settings so that CI will build the PR now

src/components/Toolbar/LinkToolbar.js
```diff
@@ -47,13 +47,15 @@ class LinkToolbar extends React.Component<Props, State> {
     this.originalValue = this.props.link.data.get("href");
     this.setState({ isEditing: !!this.originalValue });
 
-    setImmediate(() =>
-      window.addEventListener("click", this.handleOutsideMouseClick)
-    );
+    typeof window !== "undefined" &&
```
tommoor: Sorry to nit on a relatively small PR, but do you mind changing these to regular `if` statements – I think it's easier for folks to understand, particularly on these multi-line ones. But for consistency, everywhere would be great.

WNemencha: Hey, no need to apologize. We also enforce strict contributions guidelines at work, especially on making code consistent with the existing codebase. I'll fix that happily :)

tommoor:
Thanks! Released as [v9.4.0](https://github.com/outline/rich-markdown-editor/commit/2899983366545efc2cb30e646921dead3fe78203).

WNemencha:
Thanks a lot guys :)

PR #82, opened 2019-04-29
sync z-index

morloy:
I noticed that the z-index of `BlockInsert` and `CopyButton` is rather low compared to the `Toolbar`. This makes it unusable in a Modal with a z-index > 1.


tommoor:
Thanks 👍

PR #83, opened 2019-04-29
build stand-alone bundle

morloy:
Since we’re not using webpack, we need a stand-alone bundle that we can import via a `<script>` tag. Would be nice to have this bundle published on NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "lint": "yarn eslint src",
     "start": "webpack-serve --config example/webpack.config.js --content example/dist",
-    "build": "babel src/ -d lib/",
+    "build": "babel src/ -d lib/ && webpack -p",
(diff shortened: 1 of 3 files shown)


tommoor:
Hey @morloy, there shouldn't be any dependency on webpack – the compiled code is published to NPM. How are you building your application source?

morloy:
We use Meteor but can't use its import() due to our CSP. So we usually load the bundled versions (as published on CDNs) for large dependencies.

falky97:
Just wanted to clarify here that using rich-markdown-editor without a module bundler is not recommended or cannot be done?

Tried Unpkg: https://unpkg.com/rich-markdown-editor@10.2.1-0/dist/lib/index.js

Including it as a script doesn't seem to get it to work... Wondering if there are any other ways?

tommoor:
Yea, that's right – no support has been built for including directly via a script tag at this time.

PR #94, opened 2019-06-01
[CSS] Fix toolbar vertical alignment

gamerinshaft:
[before]
スクリーンショット 2019-06-02 3 34 18

<hr>

[after]
スクリーンショット 2019-06-02 3 35 38

### DIFF
--- a/src/components/Toolbar/Separator.js
+++ b/src/components/Toolbar/Separator.js
@@ -8,6 +8,7 @@ const Separator = styled.div`
   opacity: 0.2;
   display: inline-block;
   margin-left: 10px;
+  vertical-align: middle;
 `;
 
 export default Separator;


tommoor:
Hey, this isn't an issue on master for me – which browser are you testing in? It looks like this is a screenshot from the included example page – is that correct?

gamerinshaft:
I use electron `4.0.6` from [electron-react-boilerplate](https://github.com/electron-react-boilerplate/electron-react-boilerplate)
I attached screenshot about the page.

スクリーンショット 2019-06-13 10 54 02

tommoor:
Hmmm, fair. This is just very strange that changing the vertical align of the separator would make any difference to this 😆

tommoor:
I predict this is fixed by fd46b92 – unfortunately this PR isn't mergeable as it does affect the layout of the toolbar in "normal" scenarios.

Issue #93, opened 2019-05-31
"Toolbar popup" & "Block Insert Button" z-index issue with Material-ui dialog/modal


rkboyz124:
### Bug or Enhancement
Bug

### What's The Issue/Purpose
"Toolbar popup" & "Block Insert Button" is overlapped when using material-ui dialogs/modals.

### Expected Output
They should not be overlapped by the dialog/modal.

### Reproducible Demo
[Check repo here to reproduce](https://codesandbox.io/s/richtexteditor-zindex-issue-reproduction-twxso)

### Other Notes
I've seen a merged [pull-request](https://github.com/outline/rich-markdown-editor/pull/82) to fix z-index issues but the issue still persists.
I hope this one would get fixed as soon as possible.

olipo186:
Hi,

I encountered the same issue - also when trying to use the editor within a dialog/modal. The dialog simply has a higher z-index than the editor, causing the toolbar to appear behind the dialog.

The ability to configure a custom z-index would solve this issue. I submitted it as a PR: https://github.com/outline/rich-markdown-editor/pull/103

PR #103, opened 2019-07-01
feat: z-index configurable from theme

olipo186:
The hard coded z-index becomes a problem when using the editor together with other libraries, as described in this open issue for example: https://github.com/outline/rich-markdown-editor/issues/93

This PR makes the z-index configurable using the theme property. By default, all z-indices will remains the same as before to ensure backward compatibility.


tommoor:
This is great, thanks!

PR #106, opened 2019-07-31
Prevent nesting multiple links in `wrapLink` if a link is already selected

Aerlinger:
Fixes a bug where toggling a link on a link is active would cause a link node to be inserted within a link node, which results in a messy slate object structure and also produces invalid HTML.

### DIFF
--- a/src/commands.js
+++ b/src/commands.js
@@ -3,7 +3,9 @@ import { Editor, Block, KeyUtils } from "slate";
 
 const commands = {
   wrapLink(editor: Editor, href: string) {
-    editor.wrapInline({ type: "link", data: { href } });
+    if (!editor.isLinkActive()) {
+      editor.wrapInline({ type: "link", data: { href } });
+    }
   },
 
   unwrapLink(editor: Editor) {


tommoor:
Could you provide reproduction steps for the double anchor so I can double check this? Thanks

Aerlinger:
Yes. 

1. Select some text in the document.
2. Press ctrl+k to create a link
3. Select same text again and press ctrl+k

Repeating these steps causes an extraneous additional link node to be recursively wrapped in the parent, for instance, here's the document structure after doing steps 1-3:

```
{
  "object": "value",
  "document": {
    "object": "document",
    "data": {},
    "nodes": [
      {
        "object": "block",
        "type": "paragraph",
        "data": {},
        "nodes": [
          {
            "object": "text",
            "text": "",
            "marks": []
          },
          {
            "object": "inline",
            "type": "link",
            "data": {
              "href": ""
            },
            "nodes": [
              {
                "object": "text",
                "text": "",
                "marks": []
              },
              {
                "object": "inline",
                "type": "link",
                "data": {
                  "href": ""
                },
                "nodes": [
                  {
                    "object": "text",
                    "text": "test",
                    "marks": []
                  }
                ]
              },
              {
                "object": "text",
                "text": "",
                "marks": []
              }
            ]
          },
          {
            "object": "text",
            "text": "",
            "marks": []
          }
        ]
      }
    ]
  }
}
```

You should also see the following error in the console:

image

Aerlinger:
I realize this gets complicated pretty quickly when the editor selection overlaps with a link, either partially or fully as you'd have to detect when to update the existing link bounds to accomodate a new link.  In fact, as I'm typing this I'm realizing Github's editor doesn't have any specially handling for preventing nested links, though I suppose it's also not a rich text editor.

I think the ideal behavior would be something where if the editor selection is within the bounds of a link, the user would only have the option to remove/edit that link. On the other hand, if a user presses ctrl+k when the editor selection partially overlaps the link (i.e. the beginning or end of the selection falls within the link bounds) then the bounds of the original link is updated, and a new link is inserted. This is essentially what gmail and similar tools do for link editing.

Aerlinger:
An even simpler solution would be to detect if the selection covers any links at all. In that case, the ctrl+k option would do nothing (i.e. wrapLink would be a no-op) but the link toolbar would still show for the first detected link within the selection.

tommoor:
> An even simpler solution would be to detect if the selection covers any links at all. In that case, the ctrl+k option would do nothing (i.e. wrapLink would be a no-op) but the link toolbar would still show for the first detected link within the selection.

This seems to be what your solution above achieves?

Aerlinger:
Correct.

tommoor:
I'm into it 👍

PR #112, opened 2019-08-03
Fix inline markdown shortcuts being wrapped with extra space

### LINKED ISSUE #109: Extra white space inserted at end of converted inline mark after closing with spacebar
Try typing some simple inline text in the editor that includes a mark such as `**test**` and press space. After doing so an extra empty space will be inserted after the text node and before the closing `**` mark (i.e. it will serialize to `**test **` instead of `**test**`.

**Resulting editor object model:**
```js
{
  "object": "value",
  "document": {
    "object": "document",
    "data": {},
    "nodes": [
      {
        "object": "block",
        "type": "paragraph",
        "data": {},
        "nodes": [
          {
            "object": "text",
            "text": "test ",
            "marks": [
              {
                "object": "mark",
                "type": "bold",
                "data": {}
              }
            ]
          }
        ]
      }
    ]
  }
}
```

Note the empty blank space after the end of "test" in the first node.

This also shows up in the rendered markdown like so:

`**bold **`

but should instead be:

`**bold** `

**Proposed fix**:

Rather than adding a bold/italic/code/etc mark to the node during the spacebar keypress event (`onSpace`) it should probably be closed during the keypress for the closing character. In other words `**bold**` would have the "bold" mark immediately added as soon as I press the last `*`. This would also fix issues where the text doesn't get converted to a bold node if I press enter instead of space, or have any character other than a space after the mark. For instance `**test**.` won't be converted to a bold node.

Happy to submit a PR if it's worth doing so.

Aerlinger:
Fixes #109 where an extra space was being wrapped within the node with a mark. In other words pressing spacebar after typing `**bold**` would render `**bold **` instead of `**bold** `.

### DIFF
--- a/src/plugins/MarkdownShortcuts.js
+++ b/src/plugins/MarkdownShortcuts.js
@@ -90,6 +90,8 @@ export default function MarkdownShortcuts() {
       });
     }
 
+    const focusText = editor.value.focusText;
+
     for (const key of inlineShortcuts) {
       // find all inline characters
       let { mark, shortcut } = key;
@@ -97,8 +99,8 @@ export default function MarkdownShortcuts() {
 
       // only add tags if they have spaces around them or the tag is beginning
       // or the end of the block
-      for (let i = 0; i < startBlock.text.length; i++) {
-        const { text } = startBlock;
+      for (let i = 0; i < focusText.text.length; i++) {
+        const { text } = focusText;
         const start = i;
         const end = i + shortcut.length;
         const beginningOfBlock = start === 0;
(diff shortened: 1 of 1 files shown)


src/plugins/MarkdownShortcuts.js
```diff
@@ -118,16 +120,25 @@ export default function MarkdownShortcuts() {
...
+
+        ev.preventDefault();
+        // Add space character at end of the text so it doesn't get wrapped
+        // within the node containing the mark we're applying.
+        editor.insertText(' ');
+
         return editor
           .removeTextByKey(firstText.key, lastCodeTagIndex, shortcut.length)
           .removeTextByKey(firstText.key, firstCodeTagIndex, shortcut.length)
           .moveAnchorTo(firstCodeTagIndex, lastCodeTagIndex - shortcut.length)
```
Aerlinger: I'm pretty sure that moveAnchorTo only takes one argument and that this second argument isn't actually doing anything. At least I can't find any documentation for a second argument.

tommoor:
Frustratingly I just can't reproduce the original issue, the space is correctly not included in the mark. Trying to think what could possibly be different in our repro steps

tommoor:

Issue #92, opened 2019-05-27
Docs: Please add "how to hide/disable buttons"


rkboyz124:
There was a PR for [hide/disabling toolbar buttons](https://github.com/outline/rich-markdown-editor/pull/35) a few months ago which was merged, but the docs wasn't updated on how to hide/disable button selection.

chrkaatz:
Could it be, that this functionality has been disabled?

ScottAgirs:
> Could it be, that this functionality has been disabled?

Did you ever find out the answer? : )

Is this possible with `rich-markdown-editor`? How can we modify the blocks/buttons?

tommoor:
PR under way here: https://github.com/outline/rich-markdown-editor/pull/438

PR #123, opened 2019-11-13
fix: Issue #92: Add docs for hide toolbar buttons 

innomatics:
fixes https://github.com/outline/rich-markdown-editor/issues/92


tommoor:
Thanks 👍

PR #132, opened 2019-12-02
fix: Issue outline/outline#1090: Removed adding of https protocols when creating mailto links from link toolbar.

TommyJackson85:
Added an extra conditional statement for the save function, to avoid adding 'https://' to links starting with 'mailto:'.

Slight accidental change made to Link.js file, which didn't alter codes functionality.

package-lock.json has been automatically added. Unsure if it should be removed.


src/components/Toolbar/LinkToolbar.js
```diff
@@ -183,9 +183,12 @@ class LinkToolbar extends React.Component<Props, State> {
     href = href.trim();
 
     if (href) {
+      //If the input starts with 'mailto:', keep it the same.
       // If the input doesn't start with protocol or relative slash, make sure
       // a protocol is added
-      if (!href.startsWith("/") && !href.match(/^https?:\/\//i)) {
+      if (href.startsWith("mailto:")) {
+        href;
+      } else if (!href.startsWith("/") && !href.match(/^https?:\/\//i)) {
```
tommoor: I actually think the entire change in this PR could be?

```suggestion
      } else if (!href.startsWith("/") && !href.match(/^(https|mailto)?:\/\//i)) {
```

TommyJackson85:
I will change it later.

src/components/Toolbar/LinkToolbar.js
TommyJackson85: What would this 'else if' satement do? Are you just trying to ignore mailto with it? If so, it doesn't work as a single if statement. I tried. It only ignores https. 
Did you mean to also remove this code (to keep it as a single if statement)?:

`      if (href.startsWith("mailto:")) {
        href;
      }`

 or would you keep my added code? 

On a side note, break and continue were creating errors in the if function so I just recalled href instead. I know its not clean but I am open to suggestions.

TommyJackson85:
@tommoor I made a second commit and updated it to this:

`      // If the input doesn't start with mailto: or protocol or relative slash, make sure
      // a protocol is added
      if ( !href.startsWith("mailto:") && !href.startsWith("/") && !href.match(/^https?:\/\//i) ) {
        href = `https://${href}`;
      }`

I felt it was the cleanest way to do it, without creating bugs. I also removed the package-lock.json file. 

Any further suggestions would be great. Will I update my initial post?

tommoor:
Hey @TommyJackson85 please run `yarn lint --fix` to format the code so that the tests pass 😄

src/components/Toolbar/LinkToolbar.js
tommoor: Ah yes, it wouldn't work as I suggested because `mailto` links aren't in the format `mailto://` – your latest solution seems fine 👍

TommyJackson85:
Done so, committed changes and it appears to be passing.

TommyJackson85:
> Hey @TommyJackson85 please run `yarn lint --fix` to format the code so that the tests pass 😄

Did as you suggested, committed my changes and it appears to be passing now.

tommoor:
closes https://github.com/outline/outline/issues/1090

captn3m0:
@tommoor Does this work for all protocols? Or just `mailto/http(s)` URLs?

Asking because things like `steam://`, `magnet:`, `ftp://`, `tel:/` are fairly common and clickable in all browsers.

TommyJackson85:
> @tommoor Does this work for all protocols? Or just `mailto/http(s)` URLs?
> 
> Asking because things like `steam://`, `magnet:`, `ftp://`, `tel:/` are fairly common and clickable in all browsers.

@tommoor I am happy to continue working on this to fix as many protocols as possible. 
Maybe write something cleaner like:

 !href.startsWith(`mailto:` || `\`)

If we agree to this, we could rename the fix issue, create a new branch for all protocols.

tommoor:
The most elegant way would be to attempt to parse with URL protocol using the inbuilt browser methods and then fall back to the previous logic if it can't be parsed. I don't want to maintain a giant list of possible protocols in the internal logic.

TommyJackson85:
> The most elegant way would be to attempt to parse with URL protocol using the inbuilt browser methods and then fall back to the previous logic if it can't be parsed. I don't want to maintain a giant list of possible protocols in the internal logic.

something like this?

`      try {
        let url_test = new URL(href);
        console.log(url_test);
        href = url_test.href;
      } catch (e){
        if (!href.startsWith(("/")) && !href.match(/^https?:\/\//i)) {
          href = `https://${href}`;
        }
      }
`
This seams to work. I added the original code in the catch. I should note that URL automatically assumes letters or numbers before double colons, is a protocol. i.e. 'htt:' will be the detected protocol of  'htt:ps//dfdffdfdfdf'.

Will continue testing this manually in the mean time. Should I create a new pull request if I am adding any new changes?

tommoor:
> I should note that URL automatically assumes letters or numbers before double colons, is a protocol. i.e. 'htt:'

This seems fine to me, it's not possible to know all possible protocols that might be registered.

TommyJackson85:
I created a new pull request for my suggested changes because this has been merged and closed.

TommyJackson85:
> > I should note that URL automatically assumes letters or numbers before double colons, is a protocol. i.e. 'htt:'
> 
> This seems fine to me, it's not possible to know all possible protocols that might be registered on the host machine – eg custom protocols for apps on macOS.

Added creation of URL from href string in save function, to test if it has a protocol and if it is a working link. If it is a working link, href then equals the url test href. If it fails, an automatic https protocol is added if it doesn't start with one already or doesn't start with a forward slash:

try { let url_test = new URL(href); href = url_test.href; } catch (e){ if (!href.startsWith("/") && !href.match(/^https?:\/\//i)) { href =https://${href}; } }

Further development ideas for the save function: alert users if protocol is unknown even if URL creation passes. Alert users if their created link is missing a protocol, give them options from a list of common protocols, to add one.

I initial intended to include this in a new issue: 
fix: Issue outline/outline#1090: Removed adding of https protocols when creating mailto links from link toolbar. #132
I decided last minute to close the NEW issue and reopen the previous branch used for THIS issue, in which I committed my changes to it. Sorry for creating a mess.

PR #138, opened 2019-12-06
fix: save function now tests url of href string, adds https protocol if it's not working or contains forward slash.

TommyJackson85:
Added creation of URL from href string in save function, to test if it has a protocol and if it is a working link. If it is a working link, href then equals the url test href. If it fails, an automatic https protocol is added if it doesn't start with one already or doesn't start with a forward slash:

`try { let url_test = new URL(href); href = url_test.href; } catch (e){ if (!href.startsWith("/") && !href.match(/^https?:\/\//i)) { href =https://${href}; } }
`

Further development ideas for the save function: alert users if protocol is unknown even if URL creation passes. Alert users if their created link is missing a protocol, give them options from a list of common protocols, to add one.

Note: this pull request was previously closed in an attempt to reopen a [previous issue](https://github.com/outline/rich-markdown-editor/pull/132), in which this pull request tried to resolve..

### DIFF
--- a/src/components/Toolbar/LinkToolbar.js
+++ b/src/components/Toolbar/LinkToolbar.js
@@ -181,12 +181,19 @@ class LinkToolbar extends React.Component<Props, State> {
   save = (href: string) => {
     const { editor, link } = this.props;
     href = href.trim();
-
+    
     if (href) {
-      // If the input doesn't start with protocol or relative slash, make sure
-      // a protocol is added
-      if (!href.startsWith("/") && !href.match(/^https?:\/\//i)) {
-        href = `https://${href}`;
+      //If href passes the url test, 
+      //make sure href equals the new URL's href value.
+      //If it fails and if the input doesn't start with https protocol or relative slash, 
+      //make sure a default https protocol is added.
+      try {
+        let url_test = new URL(href);
+        href = url_test.href;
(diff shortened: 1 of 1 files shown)


TommyJackson85:
Reopening, pull request because I couldn't reopen original issue as it has been merged:
[fix: Issue outline/outline#1090: Removed adding of https protocols when creating mailto links from link toolbar. #132](https://github.com/outline/rich-markdown-editor/pull/132)

tommoor:
Hey @TommyJackson85 because your previous work has already been merged this PR needs to be based off of that/master. It can't be merged right now because it conflicts with the previous fix.

TommyJackson85:
> Hey @TommyJackson85 because your previous work has already been merged this PR needs to be based off of that/master. It can't be merged right now because it conflicts with the previous fix.

It's updated now with the master branch. I should note that from the catch function, I am trying to catch hrefs that already have https protocols. I don't think this is needed as a https protocol would pass the try and catch function anyway. I left it there for extra code defence.

If you are happy with this you can merge it, but if you would like to suggest any other changes I am open to continue on this.

tommoor:
I'm actually not sure if this issue is still applicable in the rewrite, however the location of the fix would likely be quite different.

PR #140, opened 2019-12-09
feat: Allow overriding individual theme properties

mush159:
Add `themeOverrides` prop to enable overriding specific properties of the theme rather than having to define an entire theme.


tommoor:

Issue #120, opened 2019-11-04
Ability to programmatically change editor value


dboris:
Is it possible to use the editor as a controlled component, or is there a way to programmatically change `editorValue`?

tommoor:
The only way currently would be to add a `key` value and force a re-mount, obviously this will lose the users cursor etc though. 

I would be happy to accept a PR that enables this ability.

TommyJackson85:
@tommoor @dboris I will  work on this.

TommyJackson85:
@tommoor something like this in what you mean by a key value? https://stackoverflow.com/questions/35792275/how-to-force-remounting-on-react-components

Would you be interested in a PR that enables an ability it allow programmatical edits aswell as human input edits at the same time?

tommoor:
Hey, yes – that's what I mean – a `key` prop. Would be open to a PR that allows programatic edits through the value prop 👍

TommyJackson85:
@tommoor please take a look at the pull request I made in relation to this issue.

I am looking for feedback from anybody as I might be going off track with the required task.

TommyJackson85:
I just need clarity on somethings. Would you of expected the key prop to be used in the src/index or the example/index?

PR #142, opened 2019-12-18
fix: Issue #120: Added auto write button to programmatically change the editor value.

TommyJackson85:
This is in no way finished but I wanted to show my progress as I will be busy the next 2 or 3 days.

To attempt to programmatically change the editor value (from [issue 120](https://github.com/outline/rich-markdown-editor/issues/120)),
I created an Auto Write button / function to re render a different Editor taking in a new deserialised string into it's value:

`
      <Editor
            id="example"
            readOnly={this.state.readOnly}
            autoWrite={this.state.autoWrite}
            value={Serializer.deserialize(autoWriteTemplate)} // autoWriteTemplate is a long string.
      />
`
but only when autoWrite is set to true.

So far it works, but I haven't set it to remount back to the original editor yet. When passing in a value into the editor it dissables the user input functionality, so it needs to set back to its original editor afterwards to allow edits.

This should be possible by resetting autoWrite to false so it re renders back, but it brings it back to its origin value, without saving the new editorValue to the autoWriteTemplate string. This is confirmed by constantly toggling auto write button (eventually I plan to have it reset back to false automatically, per click). 

I can only seem to save this value by refreshing the page when the auto write template is set. After refreshing, if I keep toggling autowrite, it shows that the template has been saved

Is there anything that needs to be called after passing the deserialised string into the value key, to make sure it is properly saved? I am still trying to understand the repository more, but so far I am learning a lot. :)

@tommoor If I am going off track with the required task here, please let me know. :) I initially set out to try and fix it from the example app first, but I've made some changes to the src/index. No rush in merging this. I am just looking for some feedback, or any suggestions here.

Further developments to do:
clean up the render function, and keep conditional if statements to a separate function to mount the correct editor (changes to the editor component are only minimum with just the addition of the value key.)

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -3,6 +3,7 @@ import * as React from "react";
 import { debounce } from "lodash";
 import ReactDOM from "react-dom";
 import Editor from "../../src";
+import Serializer from "../../src/serializer";
 
 const element = document.getElementById("main");
 const savedText = localStorage.getItem("saved");
@@ -11,6 +12,8 @@ const exampleText = `
 
 This is example content. It is persisted between reloads in localStorage.
 `;
+const autoWriteTemplate = `Document has been restarted with a template!`;
+
 const defaultValue = savedText || exampleText;
 
 class GoogleEmbed extends React.Component<*> {
@@ -23,12 +26,20 @@ class GoogleEmbed extends React.Component<*> {
 class Example extends React.Component<*, { readOnly: boolean, dark: boolean }> {
   state = {
     readOnly: false,
+    autoWrite: false,
     dark: localStorage.getItem("dark") === "enabled",
   };
+
 
   handleToggleReadOnly = () => {
     this.setState({ readOnly: !this.state.readOnly });
   };
+
+  handleToggleAutoWrite = () => {
+    if (!this.state.readOnly) {
+      this.setState({ autoWrite: !this.state.autoWrite });
+    }
+  }
 
   handleToggleDark = () => {
     const dark = !this.state.dark;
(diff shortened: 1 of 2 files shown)


tommoor:
> @tommoor If I am going off track with the required task here, please let me know. :)  I initially set out to try and fix it from the example app first, but I've made some changes to the src/index

Howdy, you'll definitely have to make changes to the index and not just the examples. I don't think we want to add a new prop here at all though, the key is just to have external changes to `value` immediately update the rendered editor.

In order to have the option of retaining the current functionality and better follow React best practices it might be nice to add a `defaultValue` prop that has the same behavior as today – only evaluating on mount.

TommyJackson85:
> 
> Howdy, you'll definitely have to make changes to the index and not just the examples. I don't think we want to add a new prop here at all though, the key is just to have external changes to `value` immediately update the rendered editor.

By props do you mean the autoWrite prop I added in src/index ? 
Am I using 'Key' correctly here? Should I set them to strings instead of numbers?


> 
> In order to have the option of retaining the current functionality and better follow React best practices it might be nice to add a `defaultValue` prop that has the same behavior as today – only evaluating on mount.

I don't understand what you mean by 'today'. defaultValue is already a prop in src/index. Do you mean keep its functionality but update it on mount if needs be? Might be a good to play around with adding defaultValue to state object in the example page.

I only worked on the example index first to see what I can and can't do with it, and to work from there.

TommyJackson85:
I am stuck at the moment with this issue. I just cant seem to understand why its not saving the added value. I took your advice on evaluated defaultValue on the component did mount. I also tried to change the state editor value and the local storage saved value, at component did mount, to try and force the change, but no luck.

`  componentDidMount() {
    if (this.props.value != undefined) {
      //setting up default value
      this.defaultValue = this.serializer.serialize(this.props.value);
      localStorage.setItem("saved", this.defaultValue);
      return this.setState({ editorValue: this.serializer.deserialize(this.defaultValue) });
    }

    this.scrollToAnchor();
    if (this.props.readOnly) return;
    if (typeof window !== "undefined") {
      window.addEventListener("keydown", this.handleKeyDown);
    }
    if (this.props.autoFocus) {
      this.focusAtEnd();
    }
  }
`

If I wasn't supposed to do this at component did mount, please let me know.
My plan was to atleast figure out how to save the changes from passing through value, before creating a remount or refreshing the page. I was able to remount from the example index.

I am open to suggestions. If you have suggestions, please give specific details on documentation for extra clarity.

If somebody wants to take over from this, they are welcome to, because I will not have a lot of time to work on this the next 7 days.

TommyJackson85:
I had to change the save the local storage from immediately clicking autowrite from the example/index. I am now looking at methods to re mount from src/index. I was considering using the a key reference again in the src/index file, but I am still trying to figure it out. I also tried this from creating a new defaultValue prop from on mount. Please check the file changes to see what I was trying to do. I am looking for any help or suggestions here. :)

tommoor:
Hey @TommyJackson85 – thanks for your efforts. I feel like the changes in this PR have ended up quite far away from what we need to fix the underlying issue.

To go back to the original problem statement, what we're trying to achieve here is simply a way to change the editor value externally, so I should be able to pass a `value` prop and when it changes the rendered editor updates.

The example doesn't necessarily need to be changed to cover this usecase as long as the documentation is.

TommyJackson85:
I didn’t mean to move away from the initial issue, but I thought it would be good to start by changing the example to switch between two example editors. 

If you would be interested, I could start a new PR, That just has a value being passed in the one example editor, with no toggling between two editors.

From what I understand about the application, is that a new value prop added to the editor will rerender the application from the src file (I noticed this from the console by adding console.logs in each function, in which render is called more than once). It renders the passed string from the value prop, showing it appear in the editor, but as a user I can not edit this without refreshing the browser in which it saves it somehow.

Now that I understand that all changes must be done from the src folder, I can make another attempt.

In regards to creating a defaultValue prop, I thought from the application it was created already from the src and this confused me. I left it out of the example editor (on auto write) to evaluate it from component on mount instead as you can see in the file changes.

tommoor:

TommyJackson85:
@tommoor no worries, I stopped working on the rich markdown editor when I noticed you were rewriting the editor. Great work so far!

PR #146, opened 2020-01-13
fix: Issue #120: Programmatically change editor value from value prop, and remounting from value prop..

TommyJackson85:
Creating a second PR on this issue, because the [initial PR](https://github.com/outline/rich-markdown-editor/pull/142) went off track and got messy. 

**As you can see from the file changes:**
On passing a value prop, a new defaultValue prop is created on mount. 
First checking for the value prop and then creating a new default value.

Then it checks if default value has been created (on a second render) and then processes the new default value through a new value prop.

It might be good to test the application without key props to see if there are differences.

**problems that need to be fixed:**

Editor still can not be manually edited after passing the Value prop into the editor. (Although I am not sure this is needed here)

From the console, a warning message appears when trying to pass this.defaultValue into the editor (from the main src folder, not the example), in which tells me I cant pass this.defaultValue as a prop. As it was suggested to me, I am trying to evaluate default value on mount. Because I can't change this.props.defaultValue, I am creating this.defaultValue instead. However, I dont really know if there is any purpose in this in that it doesn't remount the application. (I could do with further clarity on how to evaluate a new default value prop on mount. I might just be doing it wrong.)

** other issues **
Passing heading hash tags in the string that is  evaluated  from the serialiser (from the value prop) is creating loading issues. 

all passed editors could be shortened where possible to keep it clean. Maybe create a new global editor with all props except the value prop. Pass in the value prop from the render function, depending on what it needs to be. I removed some props already that aren't required.

@tommoor , any feedback would be great, as well as suggestions on the next step of solving this issue. Much of my decisions were done from your suggestions in the other PR and [posted Issue](https://github.com/outline/rich-markdown-editor/issues/120).

If anybody wants to work and improve on this, feel free to take over from me. Or just give feedback and suggestions and I will try again if needed. I have been having problems trying to solve this issue. I think I will move until something else until somebody suggests something. :)

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -3,6 +3,7 @@ import * as React from "react";
 import { debounce } from "lodash";
 import ReactDOM from "react-dom";
 import Editor from "../../src";
+import Serializer from "../../src/serializer";
 
 const element = document.getElementById("main");
 const savedText = localStorage.getItem("saved");
@@ -11,6 +12,8 @@ const exampleText = `
 
 This is example content. It is persisted between reloads in localStorage.
 `;
+const autoWriteTemplate = `Document has been restarted with a template!`;
+
 const defaultValue = savedText || exampleText;
 
 class GoogleEmbed extends React.Component<*> {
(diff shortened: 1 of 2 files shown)


ForumT:
Hi any updates on this1

TommyJackson85:
no i was waiting for feedback first.

tommoor:

Issue #131, opened 2019-12-02
Passing schema={undefined} breaks things


audiolion:
Reproducible example:

https://codesandbox.io/s/vibrant-swirles-dvc73

if you upload an image and then try to delete it, the app will break. Remove the `schema={undefined}` fixes this.

The issue appears to come from 

https://github.com/outline/rich-markdown-editor/blob/c452f73bf43f5bf0baf11169ae0d584874804ba7/src/index.js#L233-L238

where the if check on L233 passes because `null !== undefined` then on L238 we set `this.prevSchema = undefined;` and I believe having `prevSchema` as undefined breaks the Slate editor.

I realize this is a usage bug, but perhaps we can guard against this by checking if `this.props.schema` is defined before executing the schema/prevSchema updates?

TommyJackson85:
I would like to request to help you on this issue.

TommyJackson85:
@audiolion I am not finding an errors with uploading images and deleting images from my own forked repository in development, but I noticed from your codesandbox link, I am not able to delete images at all. Did you change any code for your specific project?

I added this to see what happens and because it was on the lines of what you were suggesting(was it?):
`if (this.prevSchema !== this.props.schema && this.props.schema !== undefined) {`
but it didn't allow me delete images. It did however prevent the console warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>

I should note that, when I do upload an image, it automatically changes to a default link, but I guess this is just with development and not production. (unless somebody can tell me otherwise.)

Can you give an exact scenario to what you mean?

audiolion:
@TommyJackson85, the codesandbox is a reproducible example with the exact dependencies and code that causes the bug in a minimal amount of user-land code. You should be able to copy that reproducible example to reproduce the bug. Your fork might be on a different version perhaps?

TommyJackson85:
> @TommyJackson85, the codesandbox is a reproducible example with the exact dependencies and code that causes the bug in a minimal amount of user-land code. You should be able to copy that reproducible example to reproduce the bug. Your fork might be on a different version perhaps?

Yeah but what I mean is, the app isn't crashing for me on your codesandbox. I just can't delete an image once its uploaded. Usually I click the image, its highlighted with a blue border and comman d + x deletes it (on mac). Is this the same error you are getting? Did you block clicking images somehow?

tommoor:
Thanks for filing. As of v10 the editor has been re-written in Prosemirror, so this issue is no longer applicable.

PR #148, opened 2020-01-19
fix: Issue #131: Preventing changing of schema and previous schema values when props schema is undefined.

TommyJackson85:
This pr request now prevents changing of schema and previous schema values when.. 

- props schema is undefined

Here is the Get Schema function now:

` 
getSchema = () => {
    if (this.props.schema !== undefined && this.props.schema !== this.prevSchema) {
      this.schema = {
        ...defaultSchema,
        ...(this.props.schema || {}),
      };
      this.prevSchema = this.props.schema;
    }
    return this.schema;
  };
`

Issue #131 highlighted that props schema equaling undefined can potentially break the application.

Any other suggestions to prevent other bugs from this function please mention here. I will keep updating the function and this PR in relation to suggested changes. If you think this change is going to cause other issues please also mention it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -234,7 +234,7 @@ class RichMarkdownEditor extends React.PureComponent<Props, State> {
   };
 
   getSchema = () => {
-    if (this.prevSchema !== this.props.schema) {
+    if (this.props.schema != undefined && this.props.schema !== this.prevSchema) {
       this.schema = {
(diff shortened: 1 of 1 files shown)


tommoor:
👍

TommyJackson85:
> 👍

I should of gotten this done along time ago haha! I apologise for the delay.

PR #158, opened 2020-04-05
docs: fixes onSearchLink code example position

lsei:
The `onSearchLink` code snippet was under the `onShowToast` description.

The git diff is showing this as moving the `onShowToast` description down, which is effectively the same thing.


tommoor:
Thanks for the catch

PR #161, opened 2020-04-11
fix: Removes deprecated componentWillReceiveProps

lsei:
`componentWillReceiveProps` is now deprecated and throws an error. 

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

Started by fixing it in the main toolbar component. 

I don't use tables in my setup so did not test but let me know if you notice anything funny about the toolbar.


tommoor:

Issue #179, opened 2020-05-09
Add link insert menu


tommoor:
A common usecase for Outline is to want to search for another document and insert a link including the title of the document.

Today we have the ability to highlight some text, then click on "link" and search for a document but this flow is clunky and requires writing the name of the link yourself and clicking several times.

We'd like to introduce a flow that is entirely keyboard driven like the block insert menu (perhaps an extension of it?) and that adds the link including the correct title automatically.

### To Do
Core functionality
- [x] allow cmd+k to be used without highlighting any existing text
- [x] allow document creation from the link menu as an option in the type-ahead
- [x] add some dummy search data to make development easier

Secondary
- [ ] render internal links with document icon or document emoji
- [ ] add "add link to page or url — cmd+k" item to `/` menu
- [ ] allow `/` to be used anywhere, not just a new line

Support

cc @thenanyu

thenanyu:
Some functional ideas for the workflow
1. **type-ahead for existing documents**; the simplest case is to reference a document that the user already knows exists. Similar to @mentioning workflows on github, twitter, etc
2. **JIT blank document creation**; sometimes the user may want to reference documents they intend to produce or otherwise don't yet exist. Similar to wiki media or more recently Roam Research. By typing a command and some text, the user can create a blank document with the text as the title. Or the user can highlight some text and convert that to a blank document reference. This is notably different from Notion, which converts blocks of text to documents but uses the target text as the _content_ not the title

tommoor:
This is great, 2. feels like a natural extension of 1. You type ahead, and if a matching document can't be found then there is an option to "create doc and insert link", or perhaps there is always an option to create a new doc – but at that point it's the only option.

thenanyu:
- [ ] allow `/` to be used anywhere, not just a new line
- [ ] add "add link to page or url — cmd+k" item to `/` menu
- [ ] render internal links with document icon or document emoji
- [ ] allow document creation from the link menu as an option in the type-ahead
- [ ] add some dummy search data to make development easier

PR #189, opened 2020-05-17
feat: Document create flow

tommoor:
Implements https://github.com/outline/rich-markdown-editor/issues/179

### Functionality

- [x] Autofocus input on link menu open
- [x] allow cmd+k to be used without highlighting any existing text
  - [x] raw cmd+k in the middle of an existing link should select the link
- [x] allow document creation from the link menu as an option in the type-ahead
- [x] Esc to dismiss link menu
- [x] Click outside / blur to dismiss link menu

### Debt
- [x] Update existing link editor menu to work with new callbacks
- [x] Clear previous state of link menu on close/open
- [x] Broke the caret logic somehow
- [x] DRY up Wrapper component
- [x] Restore `mark` parameter to`LinkEditor` as optional
- [x] add some dummy search data to make development easier


src/components/LinkEditor.tsx
```diff
@@ -35,14 +36,18 @@ type Props = {
...
   state = {
     selectedIndex: -1,
-    value: this.props.href,
+    value: this.href,
     results: [],
   };
 
+  get href(): string {
+    return this.props.mark ? this.props.mark.attrs.href : null;
+  }
```
thenanyu, line 49: @tommoor lmk what you think of this pattern. Consolidating the null check into a single getter

tommoor: Works for me 👍  (Although your return type should be `?string` here)

thenanyu: changed to empty string to make it play nicer with inputs

src/components/FloatingToolbar.tsx
```diff
@@ -1,170 +1,113 @@
...
-            <LinkEditor
-              mark={range.mark}
-              from={range.from}
-              to={range.to}
-              {...this.props}
-            />
-          ) : (
-            <Menu items={items} {...this.props} />
-          )}
+          {children}
```
thenanyu, line 110: - somehow this file confuses the heck out of git
- I removed the business logic and left `FloatingToolbar` as the display wrapper

src/components/SelectionToolbar.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
+import { Portal } from "react-portal";
+import { EditorView } from "prosemirror-view";
+import getTableColMenuItems from "../menus/tableCol";
+import getTableRowMenuItems from "../menus/tableRow";
+import getTableMenuItems from "../menus/table";
+import getFormattingMenuItems from "../menus/formatting";
+import FloatingToolbar from "./FloatingToolbar";
```
thenanyu, line 8: Previously, this file was the business logic from `FloatingToolbar`. Now we're moving the business logic into `SelectionToolbar` and letting `FloatingToolbar` handle the display logic

src/components/LinkMenu.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from "react";
+import { EditorView } from "prosemirror-view";
+import LinkEditor, { SearchResult } from "./LinkEditor";
+import FloatingToolbar from "./FloatingToolbar";
```
thenanyu, line 4: reusing display logic in `FloatingToolbar`

tommoor:
yo yo yo – this is ready for review + merge now?

example/src/index.js
```diff
@@ -86,15 +109,26 @@ class Example extends React.Component {
           onChange={this.handleChange}
           onClickLink={href => console.log("Clicked link: ", href)}
           onClickHashtag={tag => console.log("Clicked hashtag: ", tag)}
+          onCreateLink={title => {
+            // Delay to simulate time taken to upload
+            return new Promise((resolve, reject) => {
+              setTimeout(() => {
+                if (title !== "error") {
+                  return resolve(
+                    `https://dummyimage.com/600x400/000/fff&text=${title}`
```
tommoor, line 118: Should probably update to something more realistic

```suggestion
                    `/doc/${encodeURIComponent(title)}`
```

example/src/index.js
```diff
@@ -86,15 +109,26 @@ class Example extends React.Component {
           onChange={this.handleChange}
           onClickLink={href => console.log("Clicked link: ", href)}
           onClickHashtag={tag => console.log("Clicked hashtag: ", tag)}
+          onCreateLink={title => {
+            // Delay to simulate time taken to upload
```
tommoor, line 113:
```suggestion
            // Delay to simulate time taken for remote API request to complete
```

src/commands/createAndInsertLink.ts
```diff
@@ -0,0 +1,85 @@
...
+  return result;
+}
+
+const createAndInsertLink = async function(
+  view: EditorView,
+  title: string,
+  href: string,
+  options: {
+    onCreateLink: (title: string) => Promise<string>;
+    onShowToast?: (msg: string, code: string) => void;
```
tommoor, line 39: We're not paying by the character 😛 

```suggestion
    onShowToast?: (message: string, code: string) => void;
```

src/commands/createAndInsertLink.ts
```diff
@@ -0,0 +1,85 @@
...
+        )
+        .addMark(
+          result.pos,
+          result.pos + result.node.nodeSize,
+          state.schema.marks.link.create({ href: url })
+        )
+    );
+  } catch (err) {
+    const result = findPlaceholderLink(view.state.doc, href);
+
```
tommoor, line 66: Need to guard against the placeholder being deleted here:

```suggestion
if (!result) return;

```

src/components/LinkEditor.tsx
```diff
@@ -184,42 +227,68 @@ class LinkEditor extends React.Component<Props, State> {
...
+                icon={<DocumentIcon />}
+                onClick={this.handleSelectLink(result.url, result.title)}
+                selected={index === selectedIndex}
               />
             ))}
+
+            {showCreateLink && (
+              <LinkSearchResult
+                key="create"
+                title={`Create new doc "${value}"`}
```
tommoor, line 284:
```suggestion
                title={`Create new doc “${value}”`}
```

src/components/LinkEditor.tsx
```diff
@@ -155,25 +182,41 @@ class LinkEditor extends React.Component<Props, State> {
...
     event.preventDefault();
-    this.props.onClickLink(href);
+    this.props.onClickLink(this.href);
+  };
+
+  handleCreateLink = (value: string) => {
+    this.discardInputValue = true;
+    const { onCreateLink } = this.props;
+
+    if (onCreateLink) return onCreateLink(value);
```
tommoor, line 193: I need to add a eslint rule to enforce this

```suggestion
    if (onCreateLink) {
      return onCreateLink(value);
    }
```

src/components/LinkEditor.tsx
```diff
@@ -17,12 +23,21 @@ export type SearchResult = {
...
+  onCreateLink?: (title: string) => Promise<void>;
   onSearchLink?: (term: string) => Promise<SearchResult[]>;
+  onSelectLink: (opts: {
+    href: string;
+    title?: string;
+    from: number;
+    to: number;
+  }) => void;
   onClickLink: (url: string) => void;
+  onShowToast?: (msg: string, code: string) => void;
```
tommoor, line 40:
```suggestion
  onShowToast?: (message: string, code: string) => void;
```

src/components/LinkEditor.tsx
```diff
@@ -17,12 +23,21 @@ export type SearchResult = {
...
 type Props = {
-  mark: Mark;
+  mark?: Mark;
   from: number;
   to: number;
   tooltip: typeof React.Component;
+  onRemoveLink?: () => void;
+  onCreateLink?: (title: string) => Promise<void>;
   onSearchLink?: (term: string) => Promise<SearchResult[]>;
+  onSelectLink: (opts: {
```
tommoor, line 33:
```suggestion
  onSelectLink: (options: {
```

src/components/LinkEditor.tsx
```diff
@@ -35,14 +50,19 @@ type State = {
 
 class LinkEditor extends React.Component<Props, State> {
   discardInputValue = false;
-  initialValue: string = this.props.mark.attrs.href;
+  initialValue = this.href;
+  hasInitialSelection = this.props.from !== this.props.to;
```
tommoor, line 54: A collapsed selection is still a selection, would maybe suggest flipping this to:

```suggestion
  initialSelectionCollapsed = this.props.from === this.props.to;
```

tommoor: That sounds fine, are you going to change it to a number instead of a boolean then?

src/components/LinkEditor.tsx
tommoor: Probably not, I think a guard that just returns is more justified – don't worry about it, I'll just implement the eslint rule and automate the formatting so it's not a discussion 👍

example/src/index.js
tommoor: It would just provide a "realistic" looking document link:

image

I'm saying don't use an image url for something that's supposed to be a doc url.

src/components/LinkMenu.tsx
tommoor: Selection of "meats" appreciated

thenanyu:
changes up

tommoor:
Found a small regression in testing, I remember there was a similar issue during the original development but can't quite remember how I resolved it.

- Create a test document that's longer than the window (so it scrolls)
- Focus at the bottom
- Hit CMD+K
- The window scroll will jump to the top

tommoor:
I believe it's because the input in the floating toolbar is gaining focus before it's moved to the correct position

PR #212, opened 2020-05-29
fix: export extension and don't require onKeyDown listener

fawind:
Add two small API fixes:

* Make `Props.onKeyDown ` an optional prop by providing a no-op default callback. While the callback is useful, most users probably don't want to add their own `onKeyDown` listener.
* Export the `Extension` class. While `Props.extensions` is exposed, users can't really create their own extensions because the `Extension` class is not exported.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -64,6 +64,8 @@ import MarkdownPaste from "./plugins/MarkdownPaste";
 
 export { schema, parser, serializer } from "./server";
 
+export { default as Extension } from "./lib/Extension";
+
 export const theme = lightTheme;
 
 export type Props = {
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -112,6 +114,9 @@ class RichMarkdownEditor extends React.PureComponent<Props, State> {
     onClickLink: href => {
       window.open(href, "_blank");
     },
+    onKeyDown: () => {
```
tommoor, line 117: I think we can just define it as an optional prop in the types instead.

src/index.tsx
```diff
@@ -64,6 +64,8 @@ import MarkdownPaste from "./plugins/MarkdownPaste";
 
 export { schema, parser, serializer } from "./server";
 
+export { default as Extension } from "./lib/Extension";
```
tommoor, line 67: 👍

src/index.tsx
fawind: Updated!

PR #235, opened 2020-07-09
feat: Add optional prop to allow editing checkboxes in readOnly mode

### LINKED ISSUE #205: Add optional ability to change checkboxes when readOnly
This should be toggleable with a prop (perhaps something very explicit like `readOnlyWriteCheckboxes`) and default to off.

tommoor:
closes #205


tommoor:
Note to self: remove hard coding of node name before merging

PR #239, opened 2020-07-20
feat: Notice panels

### LINKED ISSUE #171: Add notice block type
It would be great to be able to call out pieces of content as a notice. This should be a achievable by using a named fence to serialize to.

tommoor:
New "notices" can be inserted with the shortcut ":::", or through the block menu, they look something like this:

image

New items in the block menu…

image

closes #171


src/theme.ts
```diff
@@ -50,6 +50,13 @@ export const base = {
   blockToolbarText: colors.almostBlack,
   blockToolbarHoverBackground: colors.greyLight,
   blockToolbarDivider: colors.greyMid,
+
+  noticeInfoBackground: "#F5BE31",
```
tommoor, line 54: Styling is themeable

PR #255, opened 2020-08-27
feat: add insert new line inside a table cell

### LINKED ISSUE #206: Add ability to have new lines in tables
Currently new lines are converted to spaces (stripped). Newlines can probably be encoded with `<br>`, this matches other products, although there is no accepted standard for this.

guilherme-diniz:
closes #206 

It was implemented with the command Shift+Enter, because the Ctrl+Enter & Cmd+Enter are already being used for other actions.

### DIFF
--- a/src/lib/markdown/tables.ts
+++ b/src/lib/markdown/tables.ts
@@ -12,6 +12,24 @@ export default function markdownTables(md: MarkdownIt) {
         tokens[i].level--;
       }
 
+      // convert break line into br tag
+      if (tokens[i].type === "inline" && tokens[i].content.includes("\\n")) {
+        const nodes: Token[] = [];
+        const breakParts = tokens[i].content.split(" \\n ");
+        breakParts.forEach((part, index) => {
+          const token = new Token("text", "", 1);
+          token.content = part;
+          nodes.push(token);
+
+          if (index < breakParts.length - 1) {
+            const brToken = new Token("br", "br", 1);
+            nodes.push(brToken);
+          }
+        });
+
+        tokens.splice(i, 1, ...nodes);
+      }
+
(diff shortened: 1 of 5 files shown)


src/nodes/HardBreak.ts
```diff
@@ -0,0 +1,43 @@
...
+  commands({ type }) {
+    return () => (state, dispatch) => {
+      dispatch(state.tr.replaceSelectionWith(type.create()).scrollIntoView());
+      return true;
+    };
+  }
+
+  keys({ type }) {
+    return {
+      "Shift-Enter": (state, dispatch) => {
```
tommoor, line 29: We should be able to use `Cmd` + `Enter` here, the trick is to check if we're inside a table first and only handle the event then – search the codebase for `isInTable` helper.

This will also fix an issue this has created where if you add a hard break in a regular paragraph tag it's broken when you reload.

src/lib/markdown/tables.ts
```diff
@@ -12,6 +12,24 @@ export default function markdownTables(md: MarkdownIt) {
         tokens[i].level--;
       }
 
+      // convert break line into br tag
```
tommoor, line 15: It seems there is a small bug here if you:

- add a newline into a table cell as the first item
- then add text on the second line
- reload the editor

src/nodes/HardBreak.ts
guilherme-diniz: I can't use Cmd+Enter because it's already used to save the document.
I chose Shift+Enter because it's a common shortcut to new lines in order editors.

tommoor: Yea, you're totally right.

PR #267, opened 2020-09-05
feat: Add support for custom copy via dictionary prop

joshuaslate:
Hi! Thank you for building this library. I'm evaluating using it for a side-project and realized I'd need the ability to translate the provided copy. This PR enables that by adding a `dictionary` prop.

Sample usage with i18next:

```
import React, { useMemo } from 'react';
import { useTranslation } from 'react-i18next';
import Editor from 'rich-markdown-editor';

interface Props {
  onChange: (value: string) => void;
  value: string;
}

const RichTextEditor: React.FC<Props > = ({ onChange, value }) => {
  const { t } = useTranslation('editor');

  const customDictionary = useMemo(() => ({
    addColumnAfter: t('editor:columnAfter'),
    addColumnBefore: t('editor:columnBefore'),
    addRowAfter: t('editor:rowAfter'),
    addRowBefore: t('editor:rowBefore'),
    alignCenter: t('editor:alignCenter'),
    alignLeft: t('editor:alignLeft'),
    alignRight: t('editor:alignRight'),
    bulletList: t('editor:bulletedList'),
    checkboxList: t('editor:toDoList'),
    codeBlock: t('editor:codeBlock'),
    codeCopied: t('editor:copied'),
    codeInline: t('editor:inlineCode'),
    createLink: t('editor:createLink'),
    createLinkError: t('editor:createLinkError'),
    createNewDoc: (value: string) =>t('editor:createNewDoc', { newDocName: value }),
    deleteColumn: t('editor:deleteColumn'),
    deleteRow: t('editor:deleteRow'),
    deleteTable: t('editor:deleteTable'),
    em: t('editor:em'),
    embedInvalidLink: t('editor:embedInvalidLink'),
    findOrCreateDoc: t('editor:findOrCreateDoc'),
    h1: t('editor:h1'),
    h2: t('editor:h2'),
    h3: t('editor:h3'),
    heading: t('editor:heading'),
    hr: t('editor:hr'),
    image: t('editor:image'),
    imageUploadError: t('editor:imageUploadError'),
    infoNotice: t('editor:infoNotice'),
    link: t('editor:link'),
    linkCopied: t('editor:linkCopied'),
    mark: t('editor:mark'),
    newLineEmpty: t('editor:newLineEmpty'),
    newLineWithSlash: t('editor:newLineWithSlash'),
    openLink: t('editor:openLink'),
    orderedList: t('editor:orderedList'),
    placeholder: t('editor:placeholder'),
    quote: t('editor:quote'),
    removeLink: t('editor:removeLink'),
    searchOrPasteLink: t('editor:searchOrPasteLink'),
    strikethrough: t('editor:strikethrough'),
    strong: t('editor:bold'),
    subheading: t('editor:subheading'),
    table: t('editor:table'),
    tipNotice: t('editor:tipNotice'),
    warningNotice: t('editor:warningNotice'),
  }), [t]);

  return (
    <Editor dictionary={customDictionary} defaultValue={value} onChange={onChange} />
  );
}
```


src/menus/tableRow.tsx
```diff
@@ -1,23 +1,25 @@
...
   state: EditorState,
-  index: number
+  index: number,
+  dictionary: typeof baseDictionary
 ): MenuItem[] {
   return [
     {
-      name: "addRowAfter",
-      tooltip: "Insert row before",
+      name: "addRowBefore",
```
joshuaslate, line 13: I thought maybe this one was a mistake, so I fixed it. Please let me know if my assumption was wrong.

tommoor: It's actually not a mistake, the `attrs` make it behave differently

tommoor (commented):
Hey, definitely see the value in providing i18n options… what library are you using outside the editor for that functionality?

src/menus/tableRow.tsx
joshuaslate: Thank you, reverted.

joshuaslate:
I'm using [react-i18next](https://react.i18next.com/) for i18n right now.

tommoor:
Have you seen this dictionary pattern used with other libraries? Like I say I see the value, just want to make sure that if there is a somewhat standardized pattern for components that we follow it here…

joshuaslate:
I'm not sure if there's really a standardized approach to this, but I definitely agree it makes sense not to include a full framework in the editor here.

Looking around, I see: https://www.draftail.org/docs/i18n#example, which allows consumers to pass labels directly into the blocks they're using. Because we're not manually passing blocks as props here, I don't think that approach would carry over well.

This library seems to use a similar approach to what I implemented: https://stackoverflow.com/questions/45155907/change-language-of-datepicker-in-materialize-css

I don't think a lot of the libraries out there support i18n, because a lot of apps are English-first/only.

src/index.tsx
```diff
@@ -552,6 +558,10 @@ class RichMarkdownEditor extends React.PureComponent<Props, State> {
     return this.props.theme || (this.props.dark ? darkTheme : lightTheme);
   };
 
+  dictionary = () => {
+    return { ...baseDictionary, ...this.props.dictionary };
+  };
```
tommoor, line 563: This is my only concern with the PR, as it'll create a new object every render loop it'll potentially cause sub-components to have their props have changed.

Need to have some sort of memoization here, could use lodash or getDerivedStateFromProps?

joshuaslate: True, I'll take care of that. Thanks!

tommoor:
Had to make some tweaks so merged via separate PR, but work is on master now and retains authorship 👍  – many thanks

https://github.com/outline/rich-markdown-editor/pull/270

joshuaslate:
Thank you, much appreciated!

PR #272, opened 2020-09-06
fix: hide image block when no callback is passed in

joshuaslate:
I see there is a console warning when no image upload handler is passed in, but (at least for my use case), it would be better to hide the image block when no callback is provided so the feature can be optional.


tommoor:
No problems with disabling, you'll also need a couple of checks in `nodes/Image` to disable pasting and dragging images into the editor

joshuaslate:
Thanks for the tip! I updated the MR to include the changes in `nodes/Image`.

tommoor (approved):

PR #274, opened 2020-09-07
Fix no selection floating toolbar

thenanyu:
prevent invisible selections at the end of lines
image


src/components/SelectionToolbar.tsx
```diff
@@ -34,7 +35,15 @@ function isActive(props) {
...
-  return selection && !selection.empty && !selection.node;
+  if (!selection) return false;
+  if (selection.emtpy) return false;
+  if (selection.node) return false;
+
+  const slice = selection.content();
+  const fragment = slice.content;
+  const nodes = fragment.content;
+
+  return some(nodes, n => n.content.size);
```
thenanyu, line 46: this is stupid looking but it seems ok, we only want to return the floating toolbar if there is a visible selection, meaning we're selecting things beyond empty paragraph nodes

tommoor: Without running the code, I believe there should be a proxy for this in that either slice or fragment has a size?

src/components/SelectionToolbar.tsx
```diff
@@ -34,7 +35,15 @@ function isActive(props) {
   const { view } = props;
   const { selection } = view.state;
 
-  return selection && !selection.empty && !selection.node;
+  if (!selection) return false;
+  if (selection.emtpy) return false;
```
tommoor, line 39: spelling 😬 

```suggestion
  if (selection.empty) return false;
```

src/components/SelectionToolbar.tsx
thenanyu: I couldn't find any; all the sizes come back > 0 on every level above the nodes. The thing that make me sorta nervous is the fixed hierarchy depth. I don't know if that's a stable aspect of prosemirror or if it's actually a special case that I'm hardcoding here

PR #275, opened 2020-09-07
fix: allow functional component types

joshuaslate:
TypeScript was complaining about me passing functional components `React.FC<Props>` as the icon for embeds or for tooltips. This change fixed the problem for me locally. I also noticed the second argument of `onShowToast` (the stable string/message code) wasn't included in the type definition and added it.


tommoor:
Makes sense, currently I only consume the library in a Flow codebase – moving this part to Typescript was a first step, but it makes finding these TS integration issues trickier.

PR #276, opened 2020-09-07
feat: add toast type for easier use in TypeScript

joshuaslate:
This adds a type for toast ids to make it more obvious to consumers (via autocomplete/intellisense) which values can be received in `onShowToast()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,12 +191,11 @@ The editor provides an ability to create links from the formatting toolbar for o
 />
 ```
 
-#### `onShowToast(message: string, id: ToastId): void`
+#### `onShowToast(message: string, type: ToastType): void`
 
-Triggered when the editor wishes to show an error message to the user. Hook into your apps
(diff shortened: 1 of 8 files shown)


tommoor:
Are you using this `id`? I'm wondering if it should just be removed with the v11 breaking release as the main reason for use was changing the copy which can now be done through the `dictionary` prop you added…

joshuaslate:
I am using `id` now to differentiate between different types of toasts (error, warning, info, etc.). For now, I could compare the translated strings, but if there were ever a dynamic message sent to a toast (one with a variable, like `pasteLinkWithTitle`), I couldn't do the same comparison without also having the variable. Having a consistent id seems to be the best for this use-case. What do you think?

tommoor:
An enum with `error, warning, info` values instead would actually make sense considering the usecase of this being a toast message…

joshuaslate:
I'll change it to pass an enum with those three values instead. Good call.

PR #279, opened 2020-09-10
feat: remove unnecessary bypass for URL inputs on link editor

thenanyu:
required to address https://github.com/outline/outline/issues/1527

### DIFF
--- a/src/components/LinkEditor.tsx
+++ b/src/components/LinkEditor.tsx
@@ -171,16 +171,13 @@ class LinkEditor extends React.Component<Props, State> {
 
   handleChange = async (event): Promise<void> => {
     const value = event.target.value;
-    const looksLikeUrl = isUrl(value);
 
     this.setState({
       value,
-      results: looksLikeUrl ? [] : this.state.results,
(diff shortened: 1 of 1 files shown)


src/components/LinkEditor.tsx
```diff
@@ -140,17 +140,19 @@ class LinkEditor extends React.Component<Props, State> {
...
 
       case "ArrowUp": {
+        if (event.shiftKey) return;
         event.preventDefault();
         event.stopPropagation();
         const prevIndex = this.state.selectedIndex - 1;
 
         this.setState({
-          selectedIndex: Math.max(0, prevIndex),
+          selectedIndex: Math.max(-1, prevIndex),
```
tommoor, line 149: Why change this?

thenanyu: it changes the key semantics of the search box to be similar to google (as well as many other search / type-ahead interfaces), where pressing up at the top item puts you back into the editor.

The world is about 50-50 on this behavior but the products that think about UI more intently generally do it this way IMO. It feels right to me

PR #280, opened 2020-09-10
Attachment Files : S3 Bucket

chanchadsahil7:
Background:

At Qure.ai we have been using the self-hosted outline and its feature for maintaining our internal documentation and knowledge base.
 
Problem:
We have been having a hard time uploading and maintaining files except for images on the outline that we are using.

Solution:
- So we have built a draft of the files attachment using the original image upload feature and supported all kinds of files to be upload in the outline.
- We understand that this [isssue](https://github.com/outline/outline/issues/859) has been already requested.
- The above issue requested Google Drive and Dropbox integration. However, we believe it is not a great direction to integrate third party for a feature as simple as file upload.
- S3 is a files service provider over the cloud and is already an integral part of the outline service.
- So we thought why not use the already existing S3 upload integration and support all kinds of files to be stored inside a document.

Thanks for reviewing my first open-source PR🙂


tommoor:
Hey @chanchadsahil7 – this is interesting and I'm glad you were able to fork to handle the functionality you needed. As it stands I don't think we'll be able to merge this one as it's kind of "abusing" the markdown image syntax to host files – we'd need a new way of representing files in the markdown that maintains the current image behavior.

chsasank:
> As it stands I don't think we'll be able to merge this one
 
Of course, this is only a proof of concept. We wanted to get feedback early as we develop so that the PR has higher chance of merging. We didn't want to complete the feature in a way that's not in line with your vision.

> we'd need a new way of representing files in the markdown

Indeed! How about GitHub's way of handling files as links? We can keep use link itself to differentiate the files from links render the files the slack way.

image

Markdown also allows for embedded html so that's another option.

chanchadsahil7:
Hey @tommoor thanks for the feedback but as it stands this was just for the POC and not to disturb the current markdown images syntax to host files.

- The idea is to create a similar functionality as markdown image with a file doc node in the markdown so that whenever a file is attached to the editor it creates a link to the doc uploaded.

Screenshot 2020-09-13 at 9 56 19 PM

tommoor:
> Indeed! How about GitHub's way of handling files as links? We can use some metadata to differentiate the files from links. We can also render the files the slack way.

As links makes sense for a first version both for storage in the markdown and the visual display. Id suggest adding a callback with the same signature as the image callback and only show the option in the block menu when the callback is passed in. That way users of the library can opt in to handling file uploads.

We should also provide a prop that allows setting the accepted file types

chanchadsahil7:
I am glad that we are on the same page for the first version.

Below are the points that will be taken into consideration for this feature:
- Build a file drop option on the block menu.
- Upload files to S3 bucket.
- Display files as links in the document.
- Pass a callback similar to upload the image and have a prop to validate the file upload option to be displayed inside the block menu.

@tommoor @chsasank 

I think we are good to go for the development of the first version.

chanchadsahil7:
@tommoor @chsasank 

I have pushed the code for file upload which replicates the image upload functionality for all file types.

Features added.
- Added Upload file callback.
- Added Insert all files function.
- Added a file node similar to the image node which adds files as a link in the doc.

Issues:
- File node-link getting converted to an image node-link after saving the document.

tommoor:
> File node-link getting converted to an image node-link after saving the document.

I don't think you need the `FileDoc` component at all, just insert the uploaded url as a regular link and store it as a markdown link instead of a markdown image. Would this work for your usecase?

chanchadsahil7:
> > File node-link getting converted to an image node-link after saving the document.
> 
> I don't think you need the `FileDoc` component at all, just insert the uploaded url as a regular link and store it as a markdown link instead of a markdown image. Would this work for your usecase?

Makes Sense @tommoor Thanks will make the changes.

chanchadsahil7:
@tommoor @chsasank 

I have used the markdown link for the file upload and showing in the document.

Issue:
After the file is created the href attr inside the link is wrongly parsing the URL of the file and creating a document as a draft instead of just giving a file URL as output.

tommoor:
> the href attr inside the link is wrongly parsing the URL of the file and creating a document as a draft

I'm afraid I don't know what you mean – make a gif?

chanchadsahil7:
@tommoor 

- As you can see as soon as I select the file, the file name gets integrated inside the doc and the file name is hyperlinked and the link starts with `/doc/apiattachmentsredirect` where else it should be `/api/attachments.redirect?id=b26e74cc-af3a-4dfd-a4db-6033e195b412`.

- Also when i go back to my collection I see a new document with the title as `link` created.


Upload-File-Issue

chanchadsahil7:
I think I got the issue!

> > the href attr inside the link is wrongly parsing the URL of the file and creating a document as a draft
> 
> I'm afraid I don't know what you mean – make a gif?

I think I got the fix for it.

chanchadsahil7:
Any thoughts on the PR?

example/src/index.js
```diff
@@ -166,6 +165,17 @@ class Example extends React.Component {
               );
             });
           }}
+          uploadFile={file => {
+            console.log("File upload triggered: ", file);
+
+            // Delay to simulate time taken to upload
+            return new Promise(resolve => {
+              setTimeout(
+                () => resolve("https://loremflickr.com/1000/1000"),
```
tommoor, line 174: Probably best to use a non-image for this example, otherwise it's just confusing.

src/commands/insertAllFiles.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const { state } = view;
+  const { from, to } = state.selection;
+
+  // the user might have dropped multiple files at once, we need to loop
+  for (const file of files) {
+    // start uploading the file file to the server. Using "then" syntax
+    // to allow all placeholders to be entered at once with the uploads
+    // happening in the background in parallel.
+    uploadFile(file)
```
tommoor, line 37: The delay between the user starting the upload and the files being inserted will result in strange and potentially broken scenarios – the document content can have changed substantially in that time.

With the image extension a placeholder widget is used to hold the position for this reason, but a better comparison might be the logic when creating a document from the link menu – a "placeholder" link is created and then updated when the document has been successfully created. A similar approach seems necessary here.

src/components/BlockMenu.tsx
```diff
@@ -470,6 +518,16 @@ class BlockMenu extends React.Component<Props, State> {
               />
             </VisuallyHidden>
           )}
+          {uploadFile && (
+            <VisuallyHidden>
+              <input
+                type="file"
+                ref={this.fileInputRef}
+                onChange={this.handleFilePicked}
+                accept="*"
```
tommoor, line 527: Makes sense to make this a prop

src/index.tsx
```diff
@@ -695,6 +707,21 @@ const StyledEditor = styled("div")<{
...
+  }
+
+  .file.placeholder {
+    position: relative;
+    background: ${props => props.theme.background};
+
+    a {
+      opacity: 0.5;
+    }
+  }
```
tommoor, line 723: This seems to be unused

src/lib/uploadFilePlaceholder.ts
```diff
@@ -0,0 +1,51 @@
+import { Plugin } from "prosemirror-state";
```
tommoor, line 1: The entire thing here seems to be unused now it's no longer based on an image?

src/menus/block.ts
```diff
@@ -107,6 +107,12 @@ export default function blockMenuItems(
       icon: ImageIcon,
       keywords: "picture photo",
     },
+    {
+      name: "file",
+      title: "File",
+      icon: ImageIcon,
```
tommoor, line 113: Going to need a unique icon here – there isn't a simple "file" icon in the set, but I can get one added.

src/menus/block.ts
```diff
@@ -107,6 +107,12 @@ export default function blockMenuItems(
...
+      keywords: "file doc pdf",
```
tommoor, line 114:
```suggestion
      keywords: "doc pdf",
```

No need to include "file" as the name and title are included in the keywords automatically

src/menus/block.ts
chanchadsahil7: Yes please get it added will use the same Icon here

src/menus/block.ts
chanchadsahil7: Okay got it

src/index.tsx
chanchadsahil7: sure will remove this

src/lib/uploadFilePlaceholder.ts
chanchadsahil7: sure will remove this

src/components/BlockMenu.tsx
chanchadsahil7: are you talking about the default accept prop?

src/commands/insertAllFiles.ts
chanchadsahil7: let me know if I got this right:

You are suggesting that as soon as a user selects a file using the menu option you want me to insert a place holder to the document and save it. And then start the file upload process. Once the file is uploaded just replace the link.

is that what you are suggesting?

example/src/index.js
chanchadsahil7: yep got it.

src/commands/insertAllFiles.ts
tommoor: Yes, here's the existing code:

https://github.com/outline/rich-markdown-editor/blob/7b031097489950e812f8b04e9df797b5030bcd68/src/components/LinkToolbar.tsx#L71-L88

A "placeholder" link is created with a temporary url while the document is being created. Once it's uploaded the href is swapped out. We could also target the temporary link in css to maybe fade it out or similar while the upload is happening.

marconunnari:
Hello 👋 Great work on outline and this, do we have any news about this PR?

chanchadsahil7:
Hey, I will be soon pushing the final code for the Feature.

malekhijazi:
@chanchadsahil7 any updates on this? Need any help💪  ?

warnus:
Thanks for sharing a great project.
Do you have any plans to merge this PR?

tommoor:
Honestly review was never requested again so it disappeared off my GitHub "reviews requested" list, definitely open to this.

PR #283, opened 2020-09-11
fix: Add class to floating toolbar

rblalock:
The toolbar has the potential of floating beneath things, for one example.  This allows for targeting the toolbar since there's no other way to do this it seems.

### DIFF
--- a/src/components/FloatingToolbar.tsx
+++ b/src/components/FloatingToolbar.tsx
@@ -144,6 +144,7 @@ function FloatingToolbar(props) {
   return (
     <Portal>
       <Wrapper
+        className="floating-toolbar"
         active={props.active && position.visible}
         ref={menuRef}
         offset={position.offset}


tommoor:
Hey @rblalock – we use the theme to target internal elements, random classnames are liable to go missing. Is it just the `zIndex` that you're having issues with? We could potentially add that to the theme as a more structured approach…

rblalock:
the zindex is the primarily problem yes

rblalock:
https://github.com/outline/rich-markdown-editor/blob/5bcd2371adfb363a7e57a25ec077361cbd6d9e2d/src/components/BlockMenu.tsx#L522

Hmm i guess it is possible eh? didn't know this! cool.

rblalock:
Ok I didn't realize the zIndex targets the styling of the toolbar.  That's awesome!  I better understand how that part works now so that's great.  I'm closing this since it's possible to solve my specific issue with the zIndex property in the theme

PR #284, opened 2020-09-14
fix: lodash imports to use tree shaking to reduce bundle size

ceremonious:
I've been trying to reduce our bundle size and I saw that rich-markdown-editor is importing the entirety of lodash, which is a big library. These changes should help reduce the size.

### DIFF
--- a/example/src/index.js
+++ b/example/src/index.js
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { debounce } from "lodash";
+import debounce from "lodash/debounce";
 import ReactDOM from "react-dom";
 import Editor from "../../src";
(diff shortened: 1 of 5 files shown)


thenanyu:
hey @ceremonious — I'm under the impression that the lodash maintainers recommend you do not do this, per https://lodash.com/per-method-packages

tommoor:
@thenanyu that article is referring to `lodash.pick` style, not `lodash/pick` which is okay.

I'm happy to merge this one, it hasn't been an issue in our main consuming repo because we use the babel plugin to make sure the minimum version of lodash is transpiled.

PR #298, opened 2020-09-30
fix: Cannot use multiple underscore characters within inline code mark

### LINKED ISSUE #296: Problem typing multiple underscores in identifiers
When I type multiple underscores in identifiers, for example `hello_test_hello`, the middle test part becomes italic. This even happens when I start the identifier with backticks. It seems to me that it's not possible to type identifiers like that.

**To Reproduce**
1. Type: 
 ```
`hello_test_hello`
```
2. Output = hellotesthello, with "test" part italic.

**Expected behavior**
`hello_test_hello`

**Version**
Version in getoutline.com

**Screenshots**

 - OS: Ubuntu Linux
 - Browser Chrome
 - Version Version 85.0.4183.121 (Official Build) (64-bit)

### PR DESCRIPTION
**Pull request by tommoor**
Due to the inline markdown nature of the editor we can't stop underlines from parsing entirely, but they can and should be ignored within an inline code mark.

closes #296

### DIFF
--- a/src/marks/Code.ts
+++ b/src/marks/Code.ts
@@ -30,7 +30,7 @@ export default class Code extends Mark {
 
   get schema() {
     return {
-      excludes: "strong em link mark strikethrough",
+      excludes: "_",
       parseDOM: [{ tag: "code" }],
       toDOM: () => ["code", { spellCheck: false }],
     };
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/marks/Code.ts [hunk 1]
```diff
@@ -30,7 +30,7 @@ export default class Code extends Mark {
 
   get schema() {
     return {
-      excludes: "strong em link mark strikethrough",
+      excludes: "_",
```
tommoor, line 33: Note: This is something of a coincidence – the `_` character in a prosemirror node spec means "any mark", it does not mean a literal underscore.

PR #301, opened 2020-10-06
fix: toolbar gets influenced by external styling

kuldeepk:
The external styling ends up influencing the styling of the toolbar which ends up showing buttons out of position,

Screen Shot 2020-10-06 at 10 54 21 AM

This fix neutralizes that and makes sure that buttons are not showing out of position regardless of external styling

Screen Shot 2020-10-06 at 10 54 44 AM

### DIFF
--- a/src/components/Menu.tsx
+++ b/src/components/Menu.tsx
@@ -21,7 +21,7 @@ class Menu extends React.Component<Props> {
     const Tooltip = this.props.tooltip;
 
     return (
-      <div>
+      <div style={{ display: 'flex' }}>
         {items.map((item, index) => {
           if (item.name === "separator" && item.visible !== false) {
             return <ToolbarSeparator key={index} />;


tommoor:
Hey, interested in what CSS you have that would be causing the above offset?

kuldeepk:
To clarify there isn't a particular CSS we are trying to apply. Just additional styling it was picking up from body is this,

```
body {
    margin: 0;
    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
    font-size: 1rem;
    font-weight: 400;
    line-height: 1.5;
    text-align: left;
    width: 100%;
    min-height: 100vh;
    color: #647787;
    background: #f7f8fa;
}
```

But I noticed that in some browser and some cases those buttons look out of place and noticed that it was because the toolbar is still set to `display: block`, but seems like the intention here is to set it to `display: flex` to line them up in a row. I also noticed flex values for buttons, which wouldn't take an effect unless the parent is flex. I hope this helps.

tommoor:
I can't reproduce with that css but making this flex seems fine – could you change it to using styled-components as elsewhere? Can just be called `Wrapper` or something

kuldeepk:
Sounds good. I'll push those changes.

PR #307, opened 2020-10-11
fix: Upload images with same name

### LINKED ISSUE #302: Uploading images with the same name does not work
Uploading images with the same name does not work

**To Reproduce**
1. Select "Image" in BlockMenu
2. Upload image (image.png for example)
3. Try to upload image with the same name(image.png)
**Expected behavior**
Upload two images

**Version**
All, I test on 10 and 11

Its happen because state of file input not changed
```
<input
 type="file"
 ref={this.inputRef}
 onChange={this.handleImagePicked}
 accept="image/*"
/>
```

I have solution for this bug, in BlockMenu.tsx need add before ```this.props.onClose();```
```
if (this.inputRef.current) {
  this.inputRef.current.value = "";
}
```

If this is a good solution let me know I will create PR.

aablinov:
Fix #302

### DIFF
--- a/src/components/BlockMenu.tsx
+++ b/src/components/BlockMenu.tsx
@@ -270,6 +270,10 @@ class BlockMenu extends React.Component<Props, State> {
       });
     }
 
+    if (this.inputRef.current) {
+      this.inputRef.current.value = "";
+    }
+
     this.props.onClose();
   };
 


tommoor:
Thanks for the contribution 👍

PR #312, opened 2020-10-14
Allow ability to disable menu items

jl-gogovapps:
Exposes an interface to disable items 🤷 
```
  disabledToolbarItems?: Array<string> | undefined;
  disabledBlockMenuItems?: Array<string> | undefined;
```

Certainly open to feedback here and thoughts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@gogovapps/rich-markdown-editor",
   "description": "A rich text editor with Markdown shortcuts",
-  "version": "11.0.2",
+  "version": "11.0.3-alpha1",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
   "license": "BSD-3-Clause",
--- a/package.json
+++ b/package.json
@@ -1,10 +1,11 @@
 {
-  "name": "rich-markdown-editor",
+  "name": "@gogovapps/rich-markdown-editor",
   "description": "A rich text editor with Markdown shortcuts",
   "version": "11.0.2",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
   "license": "BSD-3-Clause",
+  "publishConfig": { "registry": "https://npm.pkg.github.com/" },
   "scripts": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,10 +1,11 @@
 {
-  "name": "rich-markdown-editor",
+  "name": "@gogovapps/rich-markdown-editor",
```
jl-gogovapps, line 2: This was so I could publish it to my repo

viperfx:
Would love to have this merged in. @tommoor This would be useful.

saimonmoore:
This is an interesting feature but I'm noticing very little communication on the few PR's that are open? e.g. This PR has been open for 6 months.

tommoor:
Hi. If a feature like this were to be merged it would need to be more complete in it's removal of the individual content types - for example although this allows disabling icons in the toolbar it does not disable the associated keyboard shortcut or the markdown syntax for inserting those items.

If you need a flexible editor of this nature I would recommend https://github.com/remirror/remirror

saimonmoore:
Ok that's very good feedback indeed @tommoor .  Thanks for that and thanks for the pointer to remirror. I'm looking for a simple, flexible editor that reads and persists markdown. I'll take a good long look at it to see if it works (Still in beta but looks very promising).

viperfx:
That's really too bad, as I think most of us are looking for a simple markdown editor that is quite clean without any toolbars but has live preview similar to the outline editor.

Having a config for the menu items would be awesome - and again my use case would be to take out features to match exactly what markdown standard supports, and take out anything outline specific - rather than add anything new in. 

I hope you do consider this again @tommoor

tommoor:
> I hope you do consider this again @tommoor

Not against it in theory, just needs to be properly done to remove the entire node type from the editor rather than just the toolbar icon.

viperfx:
and just to understand you have no plans to support this internally @tommoor  - this would only possible through a community PR?

tommoor:
Thanks for the effort. This functionality was included as part of above PR ^

PR #319, opened 2020-11-08
fix: change scrollable overflow-x from scroll to auto

PhilReinking:
This PR fixes the issue where the horizontal scrollbar is always visible for tables.  

This caused, **especially in dark mode**, a visually distracting experience.

I hope the following screenshots can show the problem:

Before:
CleanShot 2020-11-08 at 17 48 33@2x

After:
CleanShot 2020-11-08 at 17 49 04@2x


tommoor:
Thanks 👍

PR #322, opened 2020-11-11
fix: Safari hangs when selecting a cell in an empty table

tommoor:
Move grip rendering to a pseudo class to work around a long standing issue with Safari/Prosemirror that causes the browser to hang in an empty table. These changes have been tested in Chrome, Safari, and Firefox.

image

closes outline/outline#1647
related ProseMirror/prosemirror#947

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1189,74 +1189,91 @@ const StyledEditor = styled("div")<{
     .selectedCell {
       background: ${props =>
         props.readOnly ? "inherit" : props.theme.tableSelectedBackground};
+
+      /* fixes Firefox background color painting over border:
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -1189,74 +1189,91 @@ const StyledEditor = styled("div")<{
     .selectedCell {
       background: ${props =>
         props.readOnly ? "inherit" : props.theme.tableSelectedBackground};
+
+      /* fixes Firefox background color painting over border:
+       * https://bugzilla.mozilla.org/show_bug.cgi?id=688556 */
+      background-clip: padding-box;
```
tommoor, line 1195: Unrelated, but I spotted this while testing Firefox.

PR #345, opened 2020-12-31
feat: Image styling based on preset classes

cliftonc:
I really want this feature, as having massive images everywhere makes our pages look really bad.  So, I took a look at the code and thought of the simplest solution.

- Potentially controversial: It re-uses the standard `title` part of the image tag to allow the user to define a class that controls the style of the image, e.g. ``.   I thought this may be OK as the project was ignoring it anyway (it wasn't in the regex) and the text is added underneath the image in a different element.
- I took a copy of the markdown-it image tag, made some minor modifications (so we can also do more later if we want to add an extra way of parsing the tag (e.g. ``) later or one of the other variants later.
- What you end up with in the markdown by default is now: `` which was the old default class of the image container div.
- If, when adding an image manually, you type `` then it will convert that into: `` in the background so that the full class is preserved at all times.  The user doesn't see this given the way the editor works.
- When the image component then renders, it renders with these classes, which have been added to the default style sheet (sorry, I'm not very good at CSS so this is very much POC!).  
- In future this could become something easily controlled by a pop-up tool bar when you click on the image, but that felt a step to far on the initial PR to see if this is a reasonable direction to take given I'm very new here :)
- I defined styles: `half`, `banner`, `small`.

Video:

https://user-images.githubusercontent.com/239305/103411844-78cd2f80-4b72-11eb-86ea-8a06a826fbd0.mov


tommoor:
This is very cool, thanks for putting it together. For easy I'll reply to your points above:

**Potentially controversial: It re-uses the standard title part of the image tag to allow the user to define a class that controls the style of the image**
I agree this is the pragmatic solution here. I originally thought that we might allow custom sizing and then store the pixel dimensions in that field, this could still be done as an addition or alternative.

**I took a copy of the markdown-it image tag, made some minor modifications**
Can you mark up your modifications? I would love to not fork that if at all possible.

**What you end up with in the markdown by default is now:  which was the old default class of the image container div.**
It'd be nice if we can avoid writing it out when it's the default case, considering the field not existing will need to be supported as default layout anyway.

**In future this could become something easily controlled by a pop-up tool bar when you click on the image**
Yep, this would need to be added before we can really close out the issue, the question becomes what are the layout options that we provide. I think "banner" might be an unneccessary one, especially given the zoom behavior is a bit awkward.

cliftonc:
Thanks for the review @tommoor.

- Removed the forked file, you are right, it wasn't needed in the end.  Initially I was going to take a path to add an extra param somewhere in the tag, but re-using title makes it work without forking.
- I now no longer write any title if there isn't one needed, and do the conversion on the fly to the class.
- I have only added `small`, `half-left` and `half-right` as these are the obvious ones.  

I'll make a start on the toolbar.

cliftonc:
One thing to add of course, is that the big negative of re-using the `"title"` field is that any one importing existing docs with it won't get what they expect - not sure if this is a showstopper (not to us as we don't use it).

tommoor:
> any one importing existing docs with it won't get what they expect

Fair, It's certainly not a field I've seen used often at all. I think it could make sense to keep an array of valid sizes and then only add the class if it's within the array, otherwise it could go on the title field. What do you think? This will also stop a potential avenue for folks to inject random classes.

cliftonc:
Works for me :) I have the toolbar mostly working, so will try and pull it all together today.

cliftonc:
Toolbar works:

https://user-images.githubusercontent.com/239305/103436098-5e4d9180-4c18-11eb-923f-99e7105acceb.mov

cliftonc:
@tommoor this is with all discussed changes in, I think the edge cases are in the styling of the `half-left` and `half-right` for different image sizes and toolbar positioning.  I am not great with CSS and so while I expect not using float, but some other flex-box technique is probably better, I don't know how to make that work without risking breaking the entire layout.

src/nodes/Image.tsx
```diff
@@ -226,51 +253,92 @@ export default class Image extends Node {
...
+      },
+      alignCenter: () => (state, dispatch) => {
+        const attrs = { ...state.selection.node.attrs, layoutClass: null };
+        dispatch(state.tr.replaceSelectionWith(type.create(attrs)));
+        return true;
+      },
     };
   }
 
+  // commands({ type }) {
```
cliftonc, line 316: I couldn't figure out what this actually did?

tommoor: This makes a `createImage` command, looks like we don't use it anywhere right now but you could leave it in there as a named command alongside the `alignX` ones you've created as it might be useful in the future.

src/nodes/Image.tsx
```diff
@@ -226,51 +253,92 @@ export default class Image extends Node {
...
+        dispatch(state.tr.deleteSelection());
+        return true;
+      },
+      alignRight: () => (state, dispatch) => {
+        const attrs = {
+          ...state.selection.node.attrs,
+          title: null,
+          layoutClass: "half-right",
+        };
+        dispatch(state.tr.replaceSelectionWith(type.create(attrs)));
```
tommoor, line 296: We might consider using `tr.setNodeMarkup` to update the attribute of the existing image node rather than replacing it entirely for these, it'll be a bit more semantically correct, there are a couple of examples you can see of that in the codebase.

src/nodes/Image.tsx
```diff
@@ -226,51 +253,92 @@ export default class Image extends Node {
...
-        state.esc(node.attrs.src) +
-        ")"
-    );
+      state.esc((node.attrs.alt || "").replace("\n", "") || "") +
+      "](" +
+      state.esc(node.attrs.src);
+    if (node.attrs.layoutClass) {
+      markdown += ' "' + state.esc(node.attrs.layoutClass) + '"';
+    }
+    if (node.attrs.title) {
```
tommoor, line 264: Should we consider if both exist? Perhaps have `layoutClass` take precedence over title with an `else`?

src/components/FloatingToolbar.tsx
```diff
@@ -82,35 +82,66 @@ function usePosition({ menuRef, isSelectingText, props }) {
...
-    left: Math.round(left + window.scrollX),
-    top: Math.round(top + window.scrollY),
-    offset: Math.round(offset),
-    visible: true,
-  };
+    const isLeftAligned = selection.node.attrs.layoutClass === "half-left";
+    const isRightAligned = selection.node.attrs.layoutClass === "half-right";
+    let left = Math.round(element.offsetLeft + element.offsetParent.offsetLeft);
+
+    if (isLeftAligned) {
```
tommoor, line 96: image

This isn't quite centred right now

src/menus/image.tsx
```diff
@@ -0,0 +1,47 @@
+import {
+  TrashIcon,
+  AlignLeftIcon,
+  AlignRightIcon,
+  AlignCenterIcon,
```
tommoor, line 5: I've now created some new icons to use for this – https://github.com/outline/outline-icons/commit/6e7bf9498a8e3271237ef7538a09750ebec02a16

You'll just need to bump the `outline-icons` dep to `v1.22.0` to get them – otherwise perfect implementation here 👌

src/index.tsx
```diff
@@ -674,27 +674,37 @@ const StyledEditor = styled("div")<{
...
     position: relative;
     background: ${props => props.theme.background};
-
     img {
       opacity: 0.5;
     }
   }
 
+  .image-half-right {
+    float: right;
```
tommoor, line 697: I'm pretty sure it's totally fine to use float, image layout was really the primary usecase for this rule.

src/nodes/Image.tsx
```diff
@@ -113,19 +126,29 @@ export default class Image extends Node {
           tag: "div[class=image]",
```
tommoor, line 126:
```suggestion
          tag: "div[class~=image]",
```

Now that the image might have those additional classes applied this selector needs updating otherwise copy/paste between documents won't work.

tommoor (commented):
Really great, confident we'll get this across the finish line quite fast 😄

src/nodes/Image.tsx
cliftonc: Ok, I'll find an example and try that.

src/menus/image.tsx
cliftonc: I added the icons - looks great, though I now get these warnings:

Screenshot 2021-01-02 at 06 50 17

src/components/FloatingToolbar.tsx
cliftonc: Thanks, I finally figured it out, I needed to traverse into the element to the actual `div.image` and now the positioning works fine and is much simpler.

src/menus/image.tsx
tommoor: Ah damn, need a better linter setup in the icons repo – I'll fix that up in a patch, they should display okay in the meantime

cliftonc:
@tommoor all fixes in as per the review, only issue is the warning I am getting in the console now from outline-icons (shown above in the comment).

src/menus/image.tsx
cliftonc: Actually I can see the issue, will send a PR to outline-icons.

src/nodes/Image.tsx
```diff
@@ -110,22 +123,32 @@ export default class Image extends Node {
       draggable: true,
       parseDOM: [
         {
-          tag: "div[class=image]",
+          tag: "div[class=~image]",
```
tommoor, line 126:
```suggestion
          tag: "div[class~=image]",
```

should be this way around

src/index.tsx
```diff
@@ -674,27 +674,37 @@ const StyledEditor = styled("div")<{
...
+    width: 50%;
+    margin-left: 2em;
+  }
+
+  .image-half-left {
+    float: left;
+    width: 50%;
+    margin-right: 2em;
```
tommoor, line 705: After testing with some real content I think we should add `margin-bottom: 1em` too otherwise paragraphs wrap too close to the caption

cliftonc:
@tommoor made those fixes and also bumped the icons.

src/nodes/Image.tsx
```diff
@@ -89,6 +83,19 @@ const uploadPlugin = options =>
     },
   });
 
+const IMAGE_CLASSES = ["half-right", "half-left"];
```
thenanyu, line 86: I think it's worth making this a little more friendly to future extension and encode with a format like `left-50`, opening up to `left-25`, `center-50` etc in the future

cliftonc: Sure, would happy to amend - thanks for reviewing.

The one risk with numbers is that they imply you can use any number and it might work (which could be a feature of course by writing the style more directly in the output).  We can also of course continue with words `quarter-left`.

What's the protocol here ?  Do we wait for @tommoor to advise ?

tommoor: I agree with Nan that this is a good idea for future extensibility. If the class was exposed to the end user I might worry about implying other options but it's not something they'll ever see so I'd be happy to just rename the ones we have 👍

cliftonc: Perfect, will rename.

tommoor:
I'd like to resolve this issue with the headings before we merge, where the meta information is not moved with the text - it also affects the "click to copy link" functionality in read-only. I think the pseudo class will need applying to a span wrapper within the heading for the float to affect it, I should be able to look at it tomorrow if you don't get to it.

image

I also found an error that happens when you try and drag and drop an image now – similarly I may get some time to investigate tomorrow:

image

cliftonc:
Ok, have had a go at fixing the styling issues, though it did change some of what was there before in terms of placement of the absolute `H1` type meta elements.  It also interfered with lists, so I had to make a change there as well.

Screenshot 2021-01-03 at 09 05 21

cliftonc:
I can't re-create the issue with drag and drop, though oddly it works fine for me in the module using `yarn start` but when I link the module into the full app I can't get drag and drop to work at all (no errors or console logs though).

I suspect the issue is here, not finding the img for some reason: https://github.com/outline/rich-markdown-editor/pull/345/files#diff-8be947e90fc37419102b633179a1850c9f78121880a1a2a5500548c8eeb4a78aR128

tommoor:
I added 2 more commits on https://github.com/outline/rich-markdown-editor/tree/fix/block-menu with fixes for the block menu positioning if you're about to pull them into your branch… They ensure the "+" button and the popover menu are positioned correctly similar to the heading markers.

image

tommoor:
Okay, I think we're good here now!

I'll give it another pass tomorrow (it's later here pacific) and hopefully merge and release a new version.

RickCogley:
What's the schedule for things like this, which are merged here, to get into the cloud version?

tommoor:
Usually within a week, in this case we found some bugs with the implementation post-merge so I didn't want to deploy the updated editor with known regressions in behavior. It'll go out once those are resolved.

cliftonc:
This one was on me @RickCogley - but I have just pushed a 'fix' that removes the left-aligned option that was causing the layout issues: https://github.com/outline/rich-markdown-editor/pull/353.

PR #346, opened 2021-01-04
feat: Add syntax highlighting support for go.

lzambarda:


tommoor:
Thanks @lzambarda – happy to add this language, not sure how it was missed – "Go" should appear before "HTML" in the list, however

lzambarda:
Hey @tommoor thanks for the quick reply. I have fixed the order in the map!

PR #351, opened 2021-01-06
fix: add spacer to fix caret on header

cliftonc:
I think the only fix I found that works is to add a spacer div (not span as the header content specifically looks for a span) so that it is at the same font size as the header text, and allow you to style it's `before` to be the small label.

Not sure if it's a hack or clever - I'll let you decide @tommoor :D

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -749,17 +749,6 @@ const StyledEditor = styled("div")<{
     margin: 1em 0 0.5em;
     font-weight: 500;
     cursor: default;
-
-    &:not(.placeholder):before {
-      display: ${props => (props.readOnly ? "none" : "inline")};
-      position: relative;
-      font-family: ${props => props.theme.fontFamilyMono};
-      color: ${props => props.theme.textSecondary};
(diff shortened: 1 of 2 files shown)


tommoor:
> Not sure if it's a hack or clever - I'll let you decide @tommoor :D

Well, it's clever but I don't think it's enough unfortunately – the horizontal alignment isn't quite right and alignment in both directions is slightly off in read-only too. Perhaps we could use the same technique as for list items and make it `position: absolute` again

cliftonc:
Ok, I'll see if I can make that work.  It also may explain why a lot of other tools (like notion) don't have the option to align and float images!

tommoor:
Perhaps, the positioning is a little unpredictable with floats unfortunately. I'm still a little uneasy about introducing the functionality because of that. If images were treated as block level elements it would help a lot – but converting them is a whole other pile of fun.

cliftonc:
Superceded by: https://github.com/outline/rich-markdown-editor/pull/353

PR #353, opened 2021-01-12
Remove left align option from images

cliftonc:
This reverts the header styles prior to the creation of the image alignment PR, as I simply found it impossible to fix the header styling in a way that worked with both copying the old functionality and also working alongside a left-aligned image.


tommoor:
Fixed instead of removed here: https://github.com/outline/rich-markdown-editor/pull/355

PR #355, opened 2021-01-13
fix: Caret size at beginning of headers

### LINKED ISSUE #350: Caret size at beginning of headers
After merging the floating images functionality (https://github.com/outline/rich-markdown-editor/pull/345) I noticed that when the cursor is placed at the beginning of a heading the caret is the wrong size, this is quite confusing and hints to the user that they won't be typing a heading.

It seems to be the difference between positioning the `::before` element relative or absolute, strangely.

cc @cliftonc

tommoor:
This relies on a "hack", introducing an invisible character between the start of the heading text and the "H1" before content so that the caret is never possible to be placed next to the "H1" ::before content.

In my testing in Chrome and Firefox this actually works really well. @cliftonc take a look, let me know what you think?

alternative to https://github.com/outline/rich-markdown-editor/pull/353
closes #350

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -751,13 +751,13 @@ const StyledEditor = styled("div")<{
     cursor: default;
 
     &:not(.placeholder):before {
-      display: ${props => (props.readOnly ? "none" : "inline")};
-      position: relative;
+      display: ${props => (props.readOnly ? "none" : "inline-block")};
       font-family: ${props => props.theme.fontFamilyMono};
       color: ${props => props.theme.textSecondary};
       font-size: 13px;
-      margin-right: -16px;
-      left: -20px;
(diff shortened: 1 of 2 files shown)


src/index.tsx
```diff
@@ -766,7 +766,12 @@ const StyledEditor = styled("div")<{
       }
     }
   }
-
+  .heading-content {
+    &:before {
+      content: "​";
```
tommoor, line 771: Note: There is an invisible character in the `content` here, I'm surprised GitHub doesn't make an exception and display it given they've been known to cause problems when misplaced.

cliftonc:
Nice!  Works for me too, and also did a quick check in Edge, and it's good 🥇

PR #356, opened 2021-01-13
fix: Allow selecting second image in a paragraph

### LINKED ISSUE #349: Image toolbar positioning
**To Reproduce**
1. Paste an image
2. Paste another image right next to it (they must be in the same paragraph)
3. Select second image
4. Toolbar appears above first image

**Expected behavior**

Should appear on the correct image

**Version**

v11.1.0

tommoor:
fix: More predictable behavior when center-aligning image

closes #349

### DIFF
--- a/src/components/FloatingToolbar.tsx
+++ b/src/components/FloatingToolbar.tsx
@@ -86,7 +86,7 @@ function usePosition({ menuRef, isSelectingText, props }) {
     selection.node && selection.node.type.name === "image";
   // Images need their own positioning to get the toolbar in the center
   if (isImageSelection) {
-    const { node: element } = view.domAtPos(selection.from);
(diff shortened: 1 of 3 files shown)


src/components/FloatingToolbar.tsx
```diff
@@ -86,7 +86,7 @@ function usePosition({ menuRef, isSelectingText, props }) {
     selection.node && selection.node.type.name === "image";
   // Images need their own positioning to get the toolbar in the center
   if (isImageSelection) {
-    const { node: element } = view.domAtPos(selection.from);
+    const element = view.nodeDOM(selection.from);
```
tommoor, line 89: This was the key to selecting the correct dom element

PR #364, opened 2021-01-26
feat: server renderToHtml method

tommoor:
Note, that this is not SSR – it's a utility method to render markdown to html using the same schema and rules as the main editor.

### TODO
(snapshots are currently wrong)

- [x] Checkbox support
- [x] Notice types should get different classes
- [x] Highlight should get a specific class
- [x] Template placeholder should get a different class


tommoor:
@thenanyu all known issues are fixed with the individual node renderers, should be nothing blocking here now 🙌

tommoor:
@thenanyu you able to push up the edits we made on your machine at some point?

tommoor:
@thenanyu cleaned up and passing

PR #370, opened 2021-01-28
Modify main marks regexes to not trigger half-way through words

lzambarda:
Proposal of regex changes to avoid toggling a style change while typing a word.
Not sure if this is too specific, but I often write variable or function names in the code and if they are in snakecase, this makes it impossible to write them due to italic constantly triggering.
Example:

this_variable_name cannot be written as it will turn into this _variable_ name (cannot replicate perfectly in github, need to surround "variable" with spaces).

I hope this makes sense!

### DIFF
--- a/src/marks/Bold.ts
+++ b/src/marks/Bold.ts
@@ -19,7 +19,7 @@ export default class Bold extends Mark {
   }
 
   inputRules({ type }) {
-    return [markInputRule(/(?:\*\*)([^*]+)(?:\*\*)$/, type)];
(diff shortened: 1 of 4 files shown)


tommoor:
Hey, thanks for submitting. I think this change would be too breaking to existing documents, can I suggest that if you're writing variable or function names you surrounded them in an inline code mark with backticks?

lzambarda:
Thanks for the reply!
Makes sense... yes I will definitely do that.

PR #378, opened 2021-02-03
Fix Interference of External Styling with Floating Toolbar Menu

dnlfrst:
As explained in #301, the floating toolbar's menu is susceptible to external styling, e.g., from [`bootstrap`](https://getbootstrap.com/) and [`bootswatch`](https://bootswatch.com/). Since I ran into the same issue and there hasn't been any progress on #301, I would like to fix the issue with this PR. As [requested](https://github.com/outline/rich-markdown-editor/pull/301#issuecomment-705288508) by @tommoor, this fix uses `styled-components`. Happy about your feedback!

### DIFF
--- a/src/components/Menu.tsx
+++ b/src/components/Menu.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { EditorView } from "prosemirror-view";
-import { withTheme } from "styled-components";
+import styled, { withTheme } from "styled-components";
 import ToolbarButton from "./ToolbarButton";
 import ToolbarSeparator from "./ToolbarSeparator";
 import theme from "../theme";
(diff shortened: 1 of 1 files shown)


tommoor:
Hi @dnlfrst – I'm happy to merge but I would like to see an example of how the css is getting overriden by bootstrap if you have a screenshot of devtools. I just don't really understand why an `unstyled` div would be affected…

dnlfrst:
Sure thing! I have already created [an MWE](https://codesandbox.io/s/mwe-for-378-of-rich-markdown-editor-3d9ks) that uses `bootstrap` and `bootswatch`. Similar to @kuldeepk, I was not able to identify the actual styling that `bootswatch` applies to cause the issue. Although I manually deactivated all stylings applied by `bootswatch` in the DevTools of Google Chrome, the issue still persisted. Obviously, it would be better to get to the root cause of this, rather than to cure its symptom.

PR #393, opened 2021-02-26
fix: Remove embed component from prosemirror attrs 

### LINKED ISSUE #392: Embeds are incompatible with YJS
Due to them passing a function as an attribute, we'll need to refactor them to enable collaborative editing via this prosemirror plugin.

tommoor:
Having a React component inside the Prosemirror attrs was not good for compatibility with other data stores, such as yjs.

closes #392
cc @aulneau


aulneau:
omg thank you! just applied these changes to my local version (forked it because i switched out emotion in place of styled-components). Works perfectly :)

tommoor:
Great, thanks for testing it out – I'll merge away.

PR #410, opened 2021-03-17
fix: Code does not highlight until editor is focused

### LINKED ISSUE #408: Highlighting doesn't apply on previous content
Hi again :~)

So i've extended the refractor/prism highlighter to include a language definition we have for the `clarity` smart contracting language, and I noticed that highlighting works fine (for any language) when you add a code block and add some code, however once you leave the document and come back to it, it seems as though prism isn't running to highlight the code. It happens with all languages, not just the custom one I've added.

I think this has something to do with the fact that I'm using yjs and the nodes/marks are affected a bit differently once it's saved/reloaded via yjs.

@tommoor I'm curious if you've ran into this with your work to make outline collaborative? I've started playing around with trying to force `prism` to highlight on mount of the editor component, but it doesn't seem to be working. Before I refactor the highlighter plugin in the editor itself, I wanted to see if there might be some guidance you'd have.

| first load | second load |
|---------|--------------|
| Screen Shot 2021-03-14 at 12 11 09 PM | Screen Shot 2021-03-14 at 12 11 17 PM |

tommoor:
closes #408 and simplifies the code while we're at it.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -265,12 +265,10 @@ class RichMarkdownEditor extends React.PureComponent<Props, State> {
         new Blockquote(),
         new CodeBlock({
           dictionary,
-          initialReadOnly: this.props.readOnly,
           onShowToast: this.props.onShowToast,
         }),
         new CodeFence({
           dictionary,
-          initialReadOnly: this.props.readOnly,
           onShowToast: this.props.onShowToast,
         }),
         new CheckboxList(),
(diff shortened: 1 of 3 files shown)


aulneau:
@tommoor just to note, I tested this in my fork and it has not solved the issue I described here https://github.com/outline/rich-markdown-editor/issues/408

tommoor:
Oh weird, well it fixed a bug that I was seeing for sure - could be yours is slightly different

PR #411, opened 2021-03-17
fix: Improve pasting behavior

### LINKED ISSUE #382: Pasting multiple links doesn't mark correctly
**To Reproduce**
1. Copy 
```
https://google.com
https://github.com
```
to the clipboard.
2. Paste the text to the editor
3. Under the hood, it transfers to
```
<https://google.com
https://github.com>
```
So the source URL is `https://google.comhttps://github.com`

**Expected behavior**
Under the hood, it transfers to
```
<https://google.com>
<https://github.com>
```

**Version**
11.2.0

tommoor:
fix: Multiple links with newlines between parsed as single link

closes #382

### DIFF
--- a/src/lib/isUrl.ts
+++ b/src/lib/isUrl.ts
@@ -1,4 +1,8 @@
 export default function isUrl(text: string) {
+  if (text.match(/\n/)) {
+    return false;
+  }
+
   try {
     new URL(text);
     return true;
--- a/src/plugins/MarkdownPaste.ts
+++ b/src/plugins/MarkdownPaste.ts
@@ -23,7 +23,8 @@ export default class MarkdownPaste extends Extension {
             const html = event.clipboardData.getData("text/html");
             const { state, dispatch } = view;
 
-            // first check if the clipboard contents can be parsed as a url
+            // first check if the clipboard contents can be parsed as a single
+            // url, this is mainly for allowing pasted urls to become embeds
             if (isUrl(text)) {
               // just paste the link mark directly onto the selected text
               if (!state.selection.empty) {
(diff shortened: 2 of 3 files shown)


src/lib/isUrl.ts
```diff
@@ -1,4 +1,8 @@
 export default function isUrl(text: string) {
+  if (text.match(/\n/)) {
```
tommoor, line 2: Seems like `new URL` strips newlines and considers whatever is left, so we need a special case here.

src/plugins/MarkdownPaste.ts
```diff
@@ -64,17 +65,23 @@ export default class MarkdownPaste extends Extension {
               return true;
             }
 
-            // otherwise, if we have html on the clipboard then fallback to the
-            // default HTML parser behavior that comes with Prosemirror.
-            if (text.length === 0 || html) return false;
+            // otherwise, if we have html on the clipboard that looks like it
+            // came from Prosemirror then use the default HTML parser behavior
+            if (text.length === 0 || (html && html.includes("data-pm-slice"))) {
```
tommoor, line 70: In reality the clipboard almost always includes html on it, whether you're pasting from the web, vscode, or a native app so the markdown parser was very rarely being used. 

We only _really_ want to use the HTML parser when copy/pasting between docs.

rune-b: Hi.
We are using rich-markdown-editor in our application and one of our use cases is pasting content from MS Word. Since Word doesn't format lists using ul/ol tags, this doesn't work out of the box, so I have written a Prosemirror plugin to intercept the pasted HTML from word (using transformPastedHTML) and replace the horrible <span> mess with proper list tags

But when upgrading from 11.4.0 to 11.6.4 my plugin doesn't work anymore. The code still runs but the result seems to be ignored by rich-markdown-editor. I'm suspecting this change as the reason, but I tried including the text 'data-pm-slice' in the html and that didn't help, so I'm not sure. 

Is there some other recent change that could have caused this change of behavior?

Here's the plugin code:
https://gist.github.com/rune-b/0a8e81ec3dc4e28446b38371050a0fe6

tommoor: Seems like a likely culprit but `data-pm-slice` not making a difference is weird in that case, as you say. Perhaps look at what the output of the paste is now it goes through the markdown formatter, it could be that your plugin can be rebuilt (maybe even smaller) based on that output.

rune-b: I did some more testing with different versions and confirmed that it stops working when going from 11.5.0 to 11.6.0, so that narrows it down a bit. Not sure what that renderToHtml commit does, could it be involved?

I don't understand what you mean by "Perhaps look at what the output of the paste is now it goes through the markdown formatter". What exactly are you suggesting here?

PR #414, opened 2021-03-20
fix: allow / to be used anywhere, not just a new line

### LINKED ISSUE #218: allow `/` to be used anywhere, not just a new line
follow-on from #179

iamsaumya:
fixes #218


iamsaumya:
Hey Tom, Please review this PR whenever you get some time. I don't have much confidence in the PR because I don't understand the codebase yet.

tommoor:
This allows the menu to be triggered but the issue is more nuanced as previously it expects the current paragraph to be empty, using the same logic on a paragraph or other node that contains content will wipe it away or attempt to insert things in invalid locations.

There just needs to be some care about ensuring that the block insertion works as expected without losing any content whereever it's called from. The hovering menu itself also needs to be placed next to the "/"

Issue #182, opened 2020-05-11
is `headingsOffset` going in the right direction?


dave-db:
10.0.0-14 again

According to the docs:

> headingsOffset
> A number that will offset the document headings by a number of levels. For example, if you already nest the editor under a main h1 title you might want the user to only be able to create h2 headings and below, in this case you would set the prop to 1.

I have that use case - all docs have an h1 set by their front matter, and the topmost heading for page content would be h2. When setting `headingsOffset` to 1 as suggested, I don't see the behaviour I was expecting.

Expected (with offset 1):

- adding "big heading" sets an h2 in markdown
- adding "medium heading" sets an h3 in markdown
- adding "small heading" sets an h4 in markdown

Actual (with offset 1)

- adding a "big heading" lists itself as an h2 in the editor (down the left), but sets an h1 in markdown
- likewise, medium and small headings are listed as h3 and h4 respectively in the block menu, but in markdown are still h2 and h3

Have I misunderstood how the prop is supposed to work?

tommoor:
The way it works right now it's only a display setting. But this was mostly because of technical restrictions in v9, in v10 it would be very easy to have the offset also affect the written markdown and with the major version jump that's a change that could be made.

Personally I don't have a usecase for the setting at this time so looking to the community for which is most likely preferred.

dave-db:
My ideal usecase would be to limit users to using h2s and below in the editor, so sign me up for that behaviour!

dave-db:
OFFTOPIC: is there a better place to ask questions about usage? I have a couple of queries relating to customising the markdown serialisation (using dashes over asterisks etc) and using image preview urls that aren't issues, and didn't want to clog things up here...

tommoor:
You can open an issue for discussion – I'll probably close to keep things clean but it provides a place that others can find it in the future and a single thread for discussion. Github has a "discussions" feature in beta that should help with this kind of thing in the future – I'm attempting to get access.

PR #415, opened 2021-03-21
fix: respect `this.options.offset` in `toMarkdown`

aulneau:
I noticed that when exporting to markdown with `headingOffset` defined, it would not adjust the level. This fixes that.

### DIFF
--- a/src/nodes/Heading.ts
+++ b/src/nodes/Heading.ts
@@ -41,26 +41,24 @@ export default class Heading extends Node {
         contentElement: "span",
       })),
       toDOM: node => {
+        const level = node.attrs.level + (this.options.offset || 0);
         const button = document.createElement("button");
         button.innerText = "#";
         button.type = "button";
         button.className = "heading-anchor";
         button.addEventListener("click", this.handleCopyLink());
 
-        return [
-          `h${node.attrs.level + (this.options.offset || 0)}`,
-          button,
-          ["span", { class: "heading-content" }, 0],
-        ];
+        return [`h${level}`, button, ["span", { class: "heading-content" }, 0]];
       },
     };
   }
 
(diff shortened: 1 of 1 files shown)


tommoor:
There's an associated issue for this here: 
https://github.com/outline/rich-markdown-editor/issues/182

I don't think the way you have the logic here will work, it will result in headings increasing in size by the offset value every time they are persisted and reloaded. The key is probably to just make this a thing that affects the UI, so selecting or using markdown `###` shortcut to create a Heading 1 actually creates a heading 2, for example. Anything that touches that offsets the actual written content will have problems

aulneau:
Ah, sorry for missing that issue.

> it will result in headings increasing in size by the offset value every time they are persisted and reloaded

Can you elaborate? It might be because my setup is different (with yjs, etc), but I've not noticed the behavior you've described. I'm also not using the read-only mode or anything like that. Is that what you're referencing? I also have some custom logic in my version that will cap the level at the levels in `this.option.levels`.

I have tested this in my editor and seems to work as expected. My use case is that I do not want the editor to allow h1's as there is a different input for the primary title of the document.

tommoor:
I was thinking something like:

- You add a `h1` in the editor
- It gets written as a `h2` with `##` in the markdown because `offset=1`
- You reload the editor
- It gets read as a `h2` from the markdown
- You save the editor
- It gets written as a `h3` with `###` in the markdown because `offset=1`

aulneau:
> It gets read as a h2 from the markdown

I see, yes, that would make sense if the behavior was such that on each load, the editor reads from a source that is markdown. With my use of the editor, the contents are never read by the editor as markdown -- it stays as prosemirror / yjs state. 

I'll close the PR :)

PR #425, opened 2021-03-31
feat: add yaml support

ecanuto:
Yaml is one of most used file format. Could be a nice to have it by default on rich-markdown-editor.

### DIFF
--- a/src/plugins/Prism.ts
+++ b/src/plugins/Prism.ts
@@ -21,6 +21,7 @@ export const LANGUAGES = {
   ruby: "Ruby",
   sql: "SQL",
   typescript: "TypeScript",
+  yaml: "Yaml",
 };
 
 type ParsedNode = {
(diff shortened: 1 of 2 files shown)


tommoor (approved):

src/plugins/Prism.ts
```diff
@@ -21,6 +21,7 @@ export const LANGUAGES = {
   ruby: "Ruby",
   sql: "SQL",
   typescript: "TypeScript",
+  yaml: "Yaml",
```
tommoor, line 24:
```suggestion
  yaml: "YAML",
```

YAML is an acronym: https://en.wikipedia.org/wiki/YAML

tommoor: If you fix this up we can land the addition

ecanuto: @tommoor change has been applied

PR #439, opened 2021-05-02
feat: add theme variables for selected block item icon and text

JordanKnott:
Currently the selected menu block item icon and text color is hard coded to be black. This adds a theme variable for both the the icon color and text color.

rte_block


tommoor:
👍 down for this in theory, good to get rid of hardcoded colors

JordanKnott:
I've ran `lint --fix` and updated the commit

tommoor (approved):

PR #441, opened 2021-05-05
fix: Improve `isUrl()` by adding RegExp

### LINKED ISSUE #440: Any text with colon in first word is treated as link
Any pasted text containing colon in the first word is treated as link

**To Reproduce**
1. Paste in some non-real-link string starting with any letters and a colon ( `abc:def some random text`)
2. See that it is treated as a link

**Expected behavior**
Treat as link only what is a link prefixed my one of common protocol prefixes and followed by a single word (`https://address.`, `mailto:address@domain.com` etc ).
Probably some RegExp would help filter them out.

**Version**
11.0.11

**Screenshots**
If applicable, add screenshots or videos to help explain your problem.

 - OS: Windows 10
 - Browser Chrome
 - Version 90

sklinov:
Improved `isURL` utility function by adding RegExp that processes `mailto:`, `http://`, `https://`, `ftp://` patterns to prevent some random strings as `abc:xyz some text` to be treated as real URL and a link.

Closes #440

### DIFF
--- a/src/lib/isUrl.ts
+++ b/src/lib/isUrl.ts
@@ -3,7 +3,10 @@ export default function isUrl(text: string) {
     return false;
   }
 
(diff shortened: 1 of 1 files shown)


tommoor:
I'd really rather not introduce a random regex when the browser has this inbuilt method for exactly the purpose. What we could do is evaluate the output and return false if `hostname === ""` perhaps?

image

sklinov:
Agree that using browser-native methods is always better, but in this case, it doesn't do all the validation needed to really tell that string is a valid URL.
Checking that `hostname === ""` would help in some cases and that would be an improvement anyway.

Should I fix my PR?

tommoor:
I think that would be better and I could merge yes, I really don't want to restrict the possible protocols to those included in a regex 👍

sklinov:
@tommoor Great! Pushed an update to PR

tommoor:
Thanks :)

PR #442, opened 2021-05-05
fix: removes image flickering after upload

jlcarvalho:
I noticed that after uploading an image there is a flicker when the in memory image is replaced by the remote image. To fix this I preloaded the remote image before the replace.

With image flickering:

https://user-images.githubusercontent.com/1238663/117144216-0e2e3980-ad88-11eb-9204-3cf7c4af1fdc.mp4

Without image flickering:

https://user-images.githubusercontent.com/1238663/117144400-4170c880-ad88-11eb-8e0e-57c516fcf393.mp4

### DIFF
--- a/src/commands/insertFiles.ts
+++ b/src/commands/insertFiles.ts
@@ -61,11 +61,15 @@ const insertFiles = function(view, event, pos, files, options) {
 
         // otherwise, insert it at the placeholder's position, and remove
         // the placeholder itself
-        const transaction = view.state.tr
-          .replaceWith(pos, pos, schema.nodes.image.create({ src }))
-          .setMeta(uploadPlaceholderPlugin, { remove: { id } });
+        const newImg = new Image();
+        newImg.onload = () => {
+          const transaction = view.state.tr
+            .replaceWith(pos, pos, schema.nodes.image.create({ src }))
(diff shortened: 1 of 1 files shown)


src/commands/insertFiles.ts
```diff
@@ -61,11 +61,15 @@ const insertFiles = function(view, event, pos, files, options) {
 
         // otherwise, insert it at the placeholder's position, and remove
         // the placeholder itself
-        const transaction = view.state.tr
-          .replaceWith(pos, pos, schema.nodes.image.create({ src }))
-          .setMeta(uploadPlaceholderPlugin, { remove: { id } });
-
-        view.dispatch(transaction);
+        const newImg = new Image();
+        newImg.onload = () => {
```
tommoor, line 65: This is a great improvement, It's been subconsciously bugging me for a while! 

We should add an `onerror` handler here that throws the err for the `catch` above, right?

jlcarvalho: Good point. Done.

https://github.com/outline/rich-markdown-editor/pull/442/commits/728fa7470d38195a8a15aad768e6481c08ec1915

tommoor (approved):

Issue #452, opened 2021-05-15
Allows intra-word emphasis using `_`


charlesharries:
Underscores within a word currently generate emphasis tags, but according to a couple of accepted Markdown specs ([GitHub-flavoured](https://github.github.com/gfm/#example-369), [CommonMark](https://spec.commonmark.org/0.29/#example-359)), they shouldn't. Crucially, however, this behaviour should only apply to `_` characters; intraword emphasis using `*` is allowed.

For example: `this_is_some_text` currently yields `this<em>is</em>some_text`, which is incorrect; it should yield `this_is_some_text`.

**Earlier work**
It looks like a sort of related issue was addressed in #298, allowing intraword underscores within code blocks. The issue here is that this behaviour doesn't catch a couple of (what seem to me) common ways of typing up code blocks (see screenshots below).

**Fix**
Getting this up to full CommonMark compliance is probably a pretty serious task, but I think that changing `Italic.inputRules()` to accept intraword underscores would be a pretty simple way to get this up and running.

I'll take a swing at this one and aim open a PR with the fix this weekend.

**To Reproduce**
1. Download the repo, install dependencies, and `yarn start` to start up Storybook
2. Navigate to **Marks** in the left sidebar.
3. After the text __a link__ in the Canvas, add a new line, and type `testing_some_text`.
4. The three underscore-separated words should merge into a single word, with the fragment "some" being italicised.

**Expected behavior**
I'd expect the three underscore-separated words to remain underscore-separated, without being formatted.

**Version**
11.8.1

**Screenshots**
![markdown](https://user-images.githubusercontent.com/20623195/118356592-e308ce80-b56d-11eb-97f7-c19a213a721d.gif)

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chromium 90/Firefox 89/Safari 14.1

**Smartphone (please complete the following information):**
 - I haven't tested on mobile but I assume that the behaviour exists here too.

PR #453, opened 2021-05-15
fix: Allow intraword underscores

charlesharries:
Fixes issue #452.

As I've said on the related issue, this doesn't make this Markdown parser _totally_ CommonMark-compliant, but it should solve a relatively common pain point writing underscore-heavy content.

I've only changed the `_` behaviour because intraword `*` _are_ allowed in Markdown.

### DIFF
--- a/src/marks/Italic.ts
+++ b/src/marks/Italic.ts
@@ -20,7 +20,7 @@ export default class Italic extends Mark {
 
   inputRules({ type }) {
     return [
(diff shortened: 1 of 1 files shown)


tommoor:
This feels good, I agree that the usecase of intraword underscores is much more likely than attempting to actually italicize part of a word (which can also be done from the formatting menu).

PR #461, opened 2021-05-29
fix: Youtube cast to boolean bug

adeboyed:
The `!!` before the url variable casts the value to a boolean, which means that the `.match` function is being called on a boolean rather than the original input string


adeboyed:
Unsure whether you'd want `!!` to the end instead, so that it obeys the type standard

PR #464, opened 2021-06-06
fix: Changing block type with existing text via slash command removes text

tommoor:
closes https://github.com/outline/outline/issues/2156

### DIFF
--- a/src/components/BlockMenu.tsx
+++ b/src/components/BlockMenu.tsx
@@ -256,18 +256,12 @@ class BlockMenu extends React.Component<Props, State> {
       onImageUploadStop,
       onShowToast,
     } = this.props;
-    const { state, dispatch } = view;
+    const { state } = view;
     const parent = findParentNode(node => !!node)(state.selection);
 
-    if (parent) {
-      dispatch(
-        state.tr.insertText(
-          "",
-          parent.pos,
-          parent.pos + parent.node.textContent.length + 1
-        )
-      );
+    this.clearSearch();
 
+    if (parent) {
       insertFiles(view, event, parent.pos, files, {
         uploadImage,
         onImageUploadStart,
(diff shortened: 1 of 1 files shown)


src/components/BlockMenu.tsx
```diff
@@ -289,13 +283,7 @@ class BlockMenu extends React.Component<Props, State> {
...
 
     if (parent) {
-      dispatch(
-        state.tr.insertText(
-          "",
-          parent.pos,
-          parent.pos + parent.node.textContent.length + 1
-        )
-      );
+      dispatch(state.tr.insertText("", parent.pos, state.selection.to));
```
tommoor, line 286: Replace just the "/slash" text rather than the entire paragraph. This does contain the assumption that the users caret is at the end. If they've used arrow keys to move the caret within the word then half will be left. Because there is no guarantee of a space character separating the search and any existing content I don't see a way around this edgecase.

RalucaRusuSNP: I managed to fix this by using `parent.start` instead of `parent.pos`. Yet, I'm not very sure if this change makes sense or not, so any feedback is welcomed.

tommoor: It's the "end" of the search word that's in question so that shouldn't make a difference from what I can tell.

accesstechnology-mike:
Brilliant thanks. Is this fix in the 0.56 release?

tommoor:
Afraid not, it'll be in `0.57`

accesstechnology-mike:
Ok, thanks! 

Does this also solve the inability to insert a /command _inline_ with existing text?

i.e. inserting a /link mid-paragraph

If it doesn't, I'll make a new request.

tommoor:
It doesn't, that's tracked as #218

Issue #490, opened 2021-07-13
table row select toolbar options miss format


JiangWeixian:
**To Reproduce**

1. clone repo & yarn & yarn start
2. go to 'table'
3. select row1 from col-`Editor` to col-`React`, toolbar will show format options, but select row1 from col-`Prosemirror` to col-`Collaborative`, toolbar only show insert options

**Expected behavior**
select full table row, toolbar should display format options + insert etc.. options

I found menu items in `menus/tableRow` not contain menuitems in `menus/formatting`, so not quite sure if it's a bug or a feature

https://github.com/outline/rich-markdown-editor/blob/9dd889932ebc688aaa1082108b24258e9b85aa1b/src/menus/tableRow.tsx#L6-L34

**Version**
11.13.0

**Screenshots**
![image](https://user-images.githubusercontent.com/6839576/125440679-2badc660-ae7d-4be9-8544-fa382858d46e.png)

![image](https://user-images.githubusercontent.com/6839576/125440737-156b888d-f0a1-4d26-9061-15d113d9a8a7.png)

**Desktop (please complete the following information):**
 - OS: macos
 - Browser chrome

PR #491, opened 2021-07-14
fix: table selection support format options

JiangWeixian:
https://github.com/outline/rich-markdown-editor/issues/490

### DIFF
--- a/src/components/SelectionToolbar.tsx
+++ b/src/components/SelectionToolbar.tsx
@@ -146,9 +146,9 @@ export default class SelectionToolbar extends React.Component<Props> {
     if (isTableSelection) {
       items = getTableMenuItems(dictionary);
     } else if (colIndex !== undefined) {
-      items = getTableColMenuItems(state, colIndex, rtl, dictionary);
+      items = getTableColMenuItems(state, colIndex, rtl, dictionary).concat(getFormattingMenuItems(state, isTemplate, dictionary));
     } else if (rowIndex !== undefined) {
-      items = getTableRowMenuItems(state, rowIndex, dictionary);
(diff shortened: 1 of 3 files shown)


tommoor:
Nice, thanks! I think we need to add a separator item between the table controls and the formatting controls to complete this

JiangWeixian:
https://github.com/outline/rich-markdown-editor/blob/9dd889932ebc688aaa1082108b24258e9b85aa1b/src/components/SelectionToolbar.tsx#L161-L163

@tommoor however, According to the above code, separator item will be filtered out.

tommoor:
Looks like that bug was recently introduced with the code for disabling extensions 🙈

tommoor:
image

Really appreciate the work here, and for finding the separator bug. Unfortunately after testing it out locally I just don't think this is going to work from a UX point of view. The toolbar is totally overwhelming and creates a bunch of extra edge cases – it's better to just select the text directly.

JiangWeixian:
> image
> 
> Really appreciate the work here, and for finding the separator bug. Unfortunately after testing it out locally I just don't think this is going to work from a UX point of view. The toolbar is totally overwhelming and creates a bunch of extra edge cases – (eg link button doesn't work) – I'm confident it's better to just select the text directly and keep the toolbars separate.

the link button is another issue, even without my pr, the link button is not working on table toolbar 😂.

JiangWeixian:
maybe is able to implement it in this way

image

tommoor:
This is a nice approach, but it does mean a reasonable amount of extra development for the extra popout positioning. 

Another related note: It would be really nice to move the add column buttons above the columns like in this screenshot, it's the same in most other tools and would remove two buttons from the toolbar then

(Also I already cherry-picked the separator fix onto `main` btw)

PR #493, opened 2021-07-15
perf: typo

JiangWeixian:
- `props.onChange` should be optional
- `style` should defined as `React.CSSProperties`

### DIFF
--- a/src/components/Flex.tsx
+++ b/src/components/Flex.tsx
@@ -16,7 +16,7 @@ type AlignValues =
   | "flex-end";
 
 type Props = {
-  style?: Record<string, string>;
+  style?: React.CSSProperties;
   column?: boolean;
   align?: AlignValues;
   justify?: JustifyValues;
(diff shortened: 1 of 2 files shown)


JiangWeixian:
image
 
🙈 there is too much lint error, may i set 

```js
"singleQuote": false
```

in `eslintrc.js`

tommoor:
eslint doesn't show these errors on my side, not sure what this would be... editor setting?

tommoor (approved):

JiangWeixian:
> eslint doesn't show these errors on my side, not sure what this would be... editor setting?

🤔 it's weird, could you execute `yarn lint` locally, if there are any lint error like these

image

tommoor:
There aren't errors, I'll also note that the default in prettier is double quotes:
https://prettier.io/docs/en/rationale.html#strings

JiangWeixian:
> There aren't errors, I'll also note that the default in prettier is double quotes:
> [prettier.io/docs/en/rationale.html#strings](https://prettier.io/docs/en/rationale.html#strings)

thx reply, i clone repo on another pc, not prettier error anymore 🎉.

PR #494, opened 2021-07-16
fix: Images in table cells are not persisted

### LINKED ISSUE #489: Images pasted into table cells don't serialize in markdown
As I understood, images can't be used in tables cells now, but they could be pasted from the clipboard into the cell and lost while serializing to markdown:
Kooha-2021-07-06-11 58 31

tommoor:
closes #489

### DIFF
--- a/src/nodes/Image.tsx
+++ b/src/nodes/Image.tsx
@@ -438,6 +438,7 @@ const Caption = styled.p`
   display: block;
   font-size: 13px;
   font-style: italic;
+  font-weight: normal;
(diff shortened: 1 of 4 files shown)


src/nodes/Image.tsx
```diff
@@ -438,6 +438,7 @@ const Caption = styled.p`
   display: block;
   font-size: 13px;
   font-style: italic;
+  font-weight: normal;
```
tommoor, line 441: fixes: Image caption is heavy weight when within table header.

PR #555, opened 2021-09-26
fix: lock gemoji to cjs version

JiangWeixian:


tommoor:
Thank you, I think this is the pragmatic solution for now.

PR #556, opened 2021-09-26
refactor: Move styles and theme to subfolder

gzuidhof:
Hi Tom and other contributors,

I'm currently integrating this editor into [Starboard Notebook](https://github.com/gzuidhof/starboard-notebook) (think Jupyter notebook but entirely in the browser), thank you for making it open source :). I will have to customize the editor a bit to get it consistent with the rest of the notebook and to get the bundle size down (e.g. by re-using the code editor for code highlighting). 

These modifications don't make sense for Outline and it's a fringe enough use-case for me to accept that the library interface will never merge such modifications - so I will be keeping most changes local. Because of the way this project is structured I can soft-fork it fairly nicely by only keeping my own modified copy of `index.tsx` with changes, the imports in that file I can still point to `rich-markdown-editor/...` which is fantastic :).

This `index.tsx` file is however pretty large, the smaller it is the less there will be to diff on changes. I noticed that the current file contains all styles in a template literal, I propose to move that to its own `styles` folder along with the `theme.ts` file - that's all this PR is!


tommoor:
I've been meaning to separate this out, great stuff

PR #559, opened 2021-10-01
Switch to lodash-es

gzuidhof:
Hi Tom,

Another small PR, this one to propose a switch to the ES Module based lodash distribution `lodash-es`.

My webpack (maybe yours too) struggles to treeshake the lodash functions, and it ends up including all 94KB of them. With this tiny change it should be able to do a better job and hopefully save every visitor 90KB.


tommoor:
Hi @gzuidhof, given that we [just had quite a nightmare](https://github.com/outline/rich-markdown-editor/pull/555) with the es version of gemoji and the lack of widespread support (having to revert it and pin to an older non-es version) we won't be able to merge this PR.

The tooling for es modules just isn't there yet and we'd immediately get folks complaining.

PR #562, opened 2021-10-03
fix: Newline creation in folded heading

### LINKED ISSUE #538: Folded headers prevent newline creation
When using folded headers on a newly added header, trying to create a new line from the folded header doesn't behave as expected. New lines are created, but they're appended to the folded content and don't show until opening. Cursor position stays on the header, and keyboard navigation breaks

**To Reproduce**
1. Create a header with some content underneath
2. Fold that header
3. Click to end of line on the header and hit enter
4. Cursor will not move, and typing will append to the current header line

**Expected behavior**
Ideally, the folded header would properly maintain its content, a new line would be created, and the cursor would be placed there (so it would behave like a normal document without the folded content). 

**Screenshots**
https://www.loom.com/share/816c1dc8b41e48a6ab49a6995801b0d5

 - OS: MacOS
 - Browser Chrome & Safari
 - RME Version 11.17.2

tommoor:
This simple looking behavior turned out to be way more code than I would have liked

closes #538

### DIFF
--- a/src/commands/splitHeading.ts
+++ b/src/commands/splitHeading.ts
@@ -1,0 +1,46 @@
+import { TextSelection } from "prosemirror-state";
+import { findBlockNodes } from "prosemirror-utils";
+import findCollapsedNodes from "../queries/findCollapsedNodes";
+
+export default function splitHeading(type) {
+  return (state, dispatch) => {
+    const { $from, from, $to, to } = state.selection;
+
+    // check we're in a matching node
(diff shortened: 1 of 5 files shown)


haveanicedavid:
> more code than I would have liked

For what it's worth, i'm glad 😁  Tried to solve this myself thinking it must be fairly simple but realized i'm not nearly familiar enough with the prosemirror API yet. I learned a lot trying though (and this PR is clarifying) so thank you for fixing!

PR #588, opened 2021-11-25
feat: add hotkey ctrl+shift+` for converting current block to notice …

RuslanGabbasov:
Our customers ask a lot of ability to set part of the text as a notice block. The currently available way to do it is to add an extra line before the text that you need to convert and select block type from /-menu.

I've added hotkey  ctrl+shift+` and one extra item to the formatting menu.

And offer you to consider this code.

### DIFF
--- a/src/menus/formatting.ts
+++ b/src/menus/formatting.ts
@@ -11,6 +11,7 @@ import {
   TodoListIcon,
   InputIcon,
   HighlightIcon,
+  InfoIcon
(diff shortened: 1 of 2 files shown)


tommoor:
Hi @RuslanGabbasov I don't think we want to add more items to the formatting menu at this time, it's honestly already too long and this command is available from the block menu.

I'd recommend keeping this change on your fork for now.

PR #589, opened 2021-11-25
feat: implemented the possibility to add extra commands into the Block menu from extensions

RuslanGabbasov:
There are a lot of cases when need to add extra commands into the block menu, like adding snippets/templates or when you try to add new elements from extended markdown specification. At the current version, we haven't any possibility for it.


RuslanGabbasov:
@tommoor Do you think this is a bad idea?

PR #590, opened 2021-11-26
feat: added syntax highlighting support for Objective-C

ltranco:


src/plugins/Prism.ts
```diff
@@ -13,6 +13,7 @@ export const LANGUAGES = {
   csharp: "C#",
   go: "Go",
   markup: "HTML",
+  objectivec: "ObjectiveC",
```
tommoor, line 16:
```suggestion
  objectivec: "Objective-C",
```

PR #594, opened 2021-11-30
fix: Emoticons converted after persistence

tommoor:
Passing the defs to match gemoji exactly may also reduce filesize as it will be using the gemoji data rather than the one bundled inside of `markdown-it-emoji`.

This also removes any chance of disconnect between these two datasets.

closes outline/outline#2785


tommoor:
cc @JiangWeixian as original author

JiangWeixian:
> cc @JiangWeixian as original author

lgtm

PR #601, opened 2021-12-14
fix: emoji highlight behavior

thenanyu:
https://www.loom.com/share/5780df97a76b487f869dedf3a1e90274

I don't know the ins and outs of prosemirror super well, but these simple config changes seem to fix the issue while behaving correctly in other conceivable ways

the best scenario would be for the emoji picker to just write the emoji text and not put a new node type into the mix but I think that would require deeper surgery


tommoor (approved):
Sweet, thank you for the video – this looks exactly right

PR #603, opened 2021-12-15
feat: Attachment of Files to Documents using S3.

FeralMib:
This is an extension of https://github.com/outline/rich-markdown-editor/pull/280 including the original commit there, rebased and extended on current master.

For further details see the original PR.

I've added support for a basic placeholder during upload, added a suitable menu icon and fixed compatibility issues during during rebasing

The matching support in outline server can be found in https://github.com/outline/outline/pull/2859


tommoor (commented):
Apart from the placeholder issue outlined in this review I think really the question is whether a plain text link is good enough – and I'm tempted to say no.

We really want to maintain a semantic link to the fact that it's a file upload within the document structure by creating a `file` node type that stores the attachmentId etc, this will allow us to do much more interesting things in the future.

cc @thenanyu

warnus:
Sorry about interrupt you guys. I really look forward to including this feature. If file upload using just link that I'm wondering that how to remove the file from storage.

